PR #30, opened 2024-02-14
fix: correct filename in blender-ffmpeg sample

zoomerkind:
the sample had the wrong filename in the inline examples.

---


ddneilson (approved):
Great! Thanks for fixing those!

PR #38, opened 2024-04-03
feat: adds a sample job template that runs scripts in docker

ddneilson:
Summary:

Prompted by a [question in the discussion forums](https://github.com/OpenJobDescription/openjd-specifications/discussions/37), this adds a very basic sample that demonstrates how one might run Tasks within a docker container.

---

### DIFF
--- a/samples/v2023-09/job_templates/bash-in-docker.yaml
+++ b/samples/v2023-09/job_templates/bash-in-docker.yaml
@@ -1,0 +1,101 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# ----
+# Demonstrates
+# ----
+# This demonstrates a way that Tasks can be run inside of a docker container.
+#
+# Run with:
+#  openjd run bash-in-docker.yaml --step RunInDocker
+#
+# ----
+# Requirements
+# ----
+# - linux (tested on Amazon Linux 2)
+# - docker (tested wth version 20.10.25)
+# - openssl (tested with version 1.0.2k-fips)
+#
+# -----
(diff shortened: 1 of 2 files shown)


samples/v2023-09/job_templates/bash-in-docker.yaml
```diff
@@ -0,0 +1,104 @@
...
+          set -xeou pipefail
+
+          # Start the docker container running in the background with a randomly determined container name
+          # The random name enabled running multiple of this same Job on the same host at the same time.
+          CONTAINER_NAME="container-$(openssl rand -hex 16)"
+          echo $CONTAINER_NAME > .docker_container_name.txt
+          nohup docker run --rm --name $CONTAINER_NAME ubuntu:24.04  /bin/bash -c "while true; do sleep 10; done" > container.out 2>&1&
+
+          # Give the container a moment to actually start
+          sleep 2
```
jusiskin: We can probably simplify this using the [`-d` (detached mode)](https://docs.docker.com/reference/cli/docker/container/run/#detach):

```suggestion
         docker run --rm -d --name $CONTAINER_NAME ubuntu:24.04  /bin/bash -c "while true; do sleep 10; done"
```

jusiskin (changes requested):
Looks good. One small suggestion to simplify starting the container in the background

samples/v2023-09/job_templates/bash-in-docker.yaml
ddneilson: Right you are. Works perfectly! Thanks

jusiskin (approved):

PR #45, opened 2024-07-08
feat(wiki): add an introduction to creating jobs

ddneilson:
This adds a guide to the wiki that walks the reader through developing a Job with Open Job Description. The guide explains elements of the Job Template as it introduces concepts bit by bit, and shows how to use the CLI to verify and test their Jobs on their workstation.

---


wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+```
+
+Success! Also verify that the output are where we expect them to be:
+
+```bash
+% ls /Users/myusername/blender_demo/output_frames 
+frame-001.png	frame-002.png
+```
+
+Before we continue adding functionality to your Job we're going to have you add a timeout to the Task. The timeout
```
ddneilson, line 506: Bug to fix, ideally before this merges: https://github.com/OpenJobDescription/openjd-cli/issues/96

wiki/Home.md
```diff
@@ -28,7 +28,8 @@ on the second talk.
 ## Getting Started
 
 The fastest way to understand the bones of Open Job Description is to understand both [How Jobs Are Constructed](How-Jobs-Are-Constructed)
-and [How Jobs Are Run](How-Jobs-Are-Run). You may also find it beneficial to look through the provided
+and [How Jobs Are Run](How-Jobs-Are-Run). Then, once you've given those a look through go through the
+[Introduction to Creating a Job](Introduction-to-Creating-a-Job) and then browse the provided
```
shidarin: Little awkward. Try

```suggestion
and [How Jobs Are Run](How-Jobs-Are-Run), followed by the [Introduction to Creating a Job](Introduction-to-Creating-a-Job) and browsing the provided
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
+# Introduction to Creating a Job
+
+This page guides you through developing a Job using Open Job Description. We'll explain elements of the Job Template bit by bit
+as we walk through creating a Job that renders an animation with Blender and then encodes those frames into a video for viewing.
+We'll also introduce you to some of the tooling that is available to help you rapidly develop a Job locally on your workstation.
+
+If you have not already read through [How Jobs Are Constructed](How-Jobs-Are-Constructed) and [How Jobs Are Run](How-Jobs-Are-Run)
+then we recommend starting with those first and then coming back once complete. Those pages provide context that will help you
```
shidarin:
```suggestion
then we recommend starting with those first. Those pages provide context that will help you
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+To replicate the steps taken in this guide you will need to have the [Open Job Description CLI](https://pypi.org/project/openjd-cli/)
+installed. This CLI provides easy to use subcommands for validating the syntax of a Job Template, running Tasks defined by a Job
+Template locally on your workstation, and more. The CLI is written in Python, so you will need to have Python 3.9 or higher
+available on your workstation; for information on how to install Python, please see the official [Python.org website](https://www.python.org/).
+
+First, we suggest creating a [Python virtual environment](https://docs.python.org/3/library/venv.html) and installing the CLI into that:
+
+| POSIX | Windows (CMD Shell) |
+| ----- | ------- | 
+| <pre>python3 -m venv .venv<br>source .venv/bin/activate<br>pip install openjd-cli</pre> | <pre>python3 -m venv .venv<br>.venv\Scripts\activate.bat<br>pip install openjd-cli</pre> |
```
shidarin: Out of scope.

```suggestion
We suggest installing the tooling into a [Python virtual environment](https://docs.python.org/3/library/venv.html).
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+available on your workstation; for information on how to install Python, please see the official [Python.org website](https://www.python.org/).
+
+First, we suggest creating a [Python virtual environment](https://docs.python.org/3/library/venv.html) and installing the CLI into that:
+
+| POSIX | Windows (CMD Shell) |
+| ----- | ------- | 
+| <pre>python3 -m venv .venv<br>source .venv/bin/activate<br>pip install openjd-cli</pre> | <pre>python3 -m venv .venv<br>.venv\Scripts\activate.bat<br>pip install openjd-cli</pre> | 
+
+For writing Job Templates by hand we also recommend using [Visual Studio Code](https://code.visualstudio.com/)
+and configuring it so that it can auto-complete the syntax of your Job Templates. To configure auto-complete, you
```
shidarin, line 59: This is so cool!!!!

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+        ]
+    }
+```
+
+To use auto-complete, simply use Control-Space and select a schema entity to automatically populate all of the entities required
+fields, or select a field to auto-complete it:
+
+| Menu | Selecting "JobTemplate" Entity | Selecting "specificationVersion" field |
+| ---  | --- | ---- |
+| ![Autocomplete menu](images/template_autocomplete_menu.png) | ![Autocomplete template](images/template_autocomplete_entity.png) | ![Autocomplete field](images/template_autocomplete_field.png) |
```
shidarin, line 94: Why are we still getting red underline errors with the specification installed?

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+and then encodes those rendered frames into an mp4 video using [FFmpeg](https://ffmpeg.org/). We illustrate 
+an approach to developing a Job that aims to keep your inner development loop fast by leveraging
+the Open Job Description CLI to test the Job on your workstation. We also demonstrates making small incremental changes to make
+it easier to narrow in on the root cause of an error. This process has faster iterations than submitting the Job
+to your distributed compute cluster, can be done in a cafe or airplane without an internet connection, and does not even require
+that you have a compute cluster that supports Open Job Description.
+
+### 2.1. Starting with shell scripts
+
+An Open Job Description Job is, essentially, describing a set of command-line commands to run, the inputs to those commands,
```
shidarin:
```suggestion
An Open Job Description Job is, essentially, describing a set of command-line instructions to run, the inputs to those commands,
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+and the order in which they need to be run. We encourage you to think of authoring a Job as writing a series of shell
+scripts that accomplish the objective of the Job when they are run in the correct order with the correct inputs.
+
+In this guide, you are creating a Job that generates frames of an animation using Blender and then encodes the frames into an animation
+using FFmpeg. You'll break this Job down into two steps:
```
shidarin:
```suggestion
In this guide, you are creating a Job that generates a sequence of frames of a Blender animation and then encodes the frames into a movie file
using FFmpeg. You'll break this Job down into two steps:
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+
+1. Generate the animation frames with Blender; and then
```
shidarin:
```suggestion
1. Generate the animation frames with Blender's command-line interface; and then
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+2. Encode the generated frames into an mp4 video with FFmpeg.
```
shidarin:
```suggestion
2. Encode the generated frames into an mp4 video file with FFmpeg.
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+
+Note that this walkthrough is written for a bash-compatible shell such as those available on typical Linux or
+MacOS workstations. If you are using Windows, then we trust that the modifications to powershell or batch from bash
+will be straightforward enough for you to handle; though you may also run bash scripts on Windows if you install
+an application such as [Git BASH](https://gitforwindows.org/).
```
shidarin:
```suggestion
MacOS workstations. If you are using Windows, you will need to modify the commands for PowerShell or batch; though you may also run bash scripts on Windows if you install
an application such as [Git BASH](https://gitforwindows.org/).
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+
+#### 2.1.1. Creating a shell script to render with Blender
+
+First, you'll create a shell script that runs Blender to render a part of the animation.
```
shidarin:
```suggestion
First, you'll create a shell script that calls the Blender command-line interface to render a part of the animation.
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+you may wish to render multiple contiguous frames of the animation in each Task. There are two ways that you can accomplish this
+with Open Job Description:
+
+1. Define Task parameters for the start and end of the range of frames to render in each Task and leverage a
+   [combination expression](2023-09-Template-Schemas#343-combinationexpr) to combine them; or
+2. Define Task parameter as a `STRING` type and hardcode the range of frames that we want to render in each Task.
+
+We'll go through both options.
+
+##### 2.2.4.2.1. With a Combination Expression
```
ddneilson, line 1186: This section needs the feature request https://github.com/OpenJobDescription/openjd-model-for-python/issues/131 to be implemented before it can be published.

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+we created the following as a starting point for our Blender script:
+
+```bash
+#!/bin/bash
+
+# Return an error code if any command in the script fails.
+set -eou pipefail
+
+# Use Blender's scripting interface to reduce the scene resolution and sampling rate to speed up testing.
+# Remove this after testing is complete.
```
shidarin:
```suggestion
# Remove this and the lines below, `  --python "$TMPFILE" \` and `rm -f $TMPFILE`, after testing is complete.
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+  --render-format PNG --use-extension 1 \
+  --render-frame "1..2"
+
+# Cleanup the temporary python script
+rm -f $TMPFILE
+```
+
+This script makes some assumptions, such as being run from a directory that contains our input and output, that we'll modify next, but
+this is a good start. It uses Blender to render two frames of the 380 frame animation created by
+the ["Barcelona Pavillion" demo scene](https://www.blender.org/download/demo-files/#cycles) (accessed July 2024) at reduced resolution.
```
shidarin:
```suggestion
the ["Barcelona Pavillion" demo scene](https://www.blender.org/download/demo-files/#cycles) by eMirage, distributed under CC-BY (accessed July 2024).
```

Reduced res has already been explained above. While we don't HAVE to attribute here (since we're not displaying the results of the file or using it directly, we should play it safe.

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+TMPFILE=$(mktemp)
+cat > "$TMPFILE" << EOF
+import bpy
+for s in bpy.data.scenes:
+  s.render.resolution_x = 480
+  s.render.resolution_y = 270
+bpy.context.scene.cycles.samples = 100
+EOF
+
+SCENE="$1"
```
shidarin, line 204: Recommend added a comment line above this which quickly explains that `$1` etc will translate into arguments on the command line calling the script, with the script itself being `$0`

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+        2. [Creating a shell script to encode a video with FFmpeg](#212-creating-a-shell-script-to-encode-a-video-with-ffmpeg)
+    2. [Creating a Job Template](#22-creating-a-job-template)
+        1. [Embed the scripts into a Job Template](#221-embed-the-scripts-into-a-job-template)
+        2. [Parameterizing the template](#222-parameterizing-the-template)
+        3. [Path Mapping](#223-path-mapping)
+        4. [Adding Task Parallelism](#224-adding-task-parallelism)
+    3. [Ready for Production](#23-ready-for-production)
+3. [Next Steps](#3-next-steps)
+
+## 1. Toolchain Setup
```
shidarin, line 47: You need to mention that it requires Blender and FFmpeg installed as well.

shidarin: I would fix this before merging

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+
+Then, as with the Blender script, save it to a file (`encode.sh`), set its execute bits, and then test it:
+
+```bash
+% ./encode.sh 
+ffmpeg version 6.1.1-tessus  https://evermeet.cx/ffmpeg/  Copyright (c) 2000-2023 the FFmpeg developers
+...
+```
+
+Then modify it so that it accepts script arguments:
```
shidarin: script arguments or CLI arguments?

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+
+## 2. Walkthrough
+
+To demonstrate one way that you could develop a Job for Open Job Description this walkthrough will guide
+you through creating a Job Template that generates frames of an animation using [Blender](https://www.blender.org/),
+and then encodes those rendered frames into an mp4 video using [FFmpeg](https://ffmpeg.org/). We illustrate 
+an approach to developing a Job that aims to keep your inner development loop fast by leveraging
+the Open Job Description CLI to test the Job on your workstation. We also demonstrates making small incremental changes to make
+it easier to narrow in on the root cause of an error. This process has faster iterations than submitting the Job
+to your distributed compute cluster, can be done in a cafe or airplane without an internet connection, and does not even require
```
shidarin:
```suggestion
to your distributed compute cluster (AKA render farm), can be done in a cafe or airplane without an internet connection, and does not even require
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+        onRun:
+          command: 
+```
+
+The way to think about a Step in Open Job Description is that it is defining a command to run and the collection of inputs
+to run the command with. This is a [map operation](https://en.wikipedia.org/wiki/Map_(higher-order_function)), if
+you are familiar with that concept. You're not going to define any inputs for the command yet, so you'll be using the default
+of having a single Task with empty input. The command is defined in the `onRun` Action in the skeleton above; this is
+the command that is run when running a Task within a [Session](How-Jobs-Are-Run#sessions). An Action in the 2023-09 
+revision of the specification [is defined as](2023-09-Template-Schemas#5-action):
```
shidarin: We suddenly up the reading and tech level a lot with this paragraph and the few that follow. Consider shorter sentences and more of them.

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+args: [ <ArgString>, ... ] # @optional @fmtstring[host]
+timeout: <posinteger> # @optional
+cancelation: <CancelationMethod> # @optional
+```
+
+Notice that both the `command` and elements of the `args` array are annotated with `@fmtstring[host]` indicating that
+they are [Format Strings](How-Jobs-Are-Constructed#format-strings) that are evaluated on the host that the command will
+run on. We're going to make use of this property to create our Job.
+
+You've already expressed your Job as a series of two shell scripts to run in sequence, so next you'll make use of the
```
shidarin: Feels like we skipped a step- we haven't had the OpenJD example of using the script itself yet, so "you've already expressed your job as.." doesn't make much sense. We've only been running single scripts, not any Jobs yet

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+name: DemoJob
+steps:
+  - name: BlenderRender
+    script:
+      actions:
+        onRun:
+          # Note: Task.File.Render refers to the file location on disk where the
+          #  contents of the embedded file named "Render" are materialized.
+          command: "{{Task.File.Render}}"
+          args:
```
shidarin, line 364: Be good to comment here that this *needs to be a list* as it's one area we've been people butt their heads against (including me)

shidarin (commented):
Mostly nits, couple rewords, some details missing.

wiki/Introduction-to-Creating-a-Job.md
ddneilson: Mostly that the values are blank, so vs-code is highlighting that. The highlight on the standalone specificationVersion line is because there are other required fields that aren't yet present.

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+Mon Jul  8 10:48:30 2024	
+Mon Jul  8 10:48:30 2024	==============================================
+Mon Jul  8 10:48:30 2024	--------- Running Task
+Mon Jul  8 10:48:30 2024	==============================================
+Mon Jul  8 10:48:30 2024	Parameter values:
+Mon Jul  8 10:48:30 2024	Frames(STRING) = 1..40
+...
+```
+
+### 2.3. Ready for Production
```
ddneilson, line 1581: Note to self: We should add something on the `hostRequirements` property in here.

wiki/Introduction-to-Creating-a-Job.md
shidarin: Coming back with fresh eyes, I would phrase it as "We've created two scripts and can manually run them in sequence to approximate the results we want. We can use Open Job Description's concept of embedded files to put those scripts directly into a Job template, which..

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+          type: TEXT
+          filename: render.sh
+          runnable: true
+          data: |
+            #!/bin/bash
+
+            # Return an error code if any command in the script fails.
+            set -eou pipefail
+
+            TMPFILE=$(mktemp)
```
shidarin, line 383: This is missing the comments from the above

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+            blender --background "$SCENE" \
+              --python "$TMPFILE" \
+              --render-output "$OUTDIR"/frame-### \
+              --render-format PNG --use-extension 1 \
+              --render-frame "${START_FRAME}..${END_FRAME}"
+
+            rm -f $TMPFILE
+```
+
+Note that if you are writing your Job for Windows with the render script being a batch or powershell script, then you will
```
shidarin:
```suggestion
Note that if you are writing your Job for Windows with the render script being a batch or PowerShell script, then you will
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+  - "{{Task.File.Render}}"
+  - 3d/pavillon_barcelone_v1.2.blend
+  - output_frames
+  - "1"
+  - "2"
+```
+
+Next, following our incremental approach, test what you have so far locally on your workstation using the
+Open Job Description CLI (reminder: activate your Python virtual environment first, if you are using one). First,
+make sure that you do not have any syntax errors in our template using the CLI's `check` command:
```
shidarin:
```suggestion
Before moving on, it's a good idea to check to the syntax of the file. Make sure to activate any Python virtual env as needed, then use the CLI's `check` command:
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+Fri Jul  5 09:33:02 2024	Command started as pid: 85706
+Fri Jul  5 09:33:02 2024	Output:
+Fri Jul  5 09:33:03 2024	Blender 4.1.1 (hash e1743a0317bc built 2024-04-16 00:06:22)
+Fri Jul  5 09:33:03 2024	Error: Cannot read file "/private/var/folders/93/p7mk60kn69ggyc87dfbwtht80000gr/T/OpenJD/sample_session0bd5l1au/3d/pavillon_barcelone_v1.2.blend": No such file or directory
+Fri Jul  5 09:33:03 2024	
+Fri Jul  5 09:33:03 2024	Blender quit
+...
+```
+
+Uh-oh, there's an error! Blender couldn't find the input file. This is because all actions in a Session are always run with their
```
shidarin: Nit: "The error means that Blender couldn't find the input file."

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+% openjd run --step BlenderRender job.template.yaml
+Fri Jul  5 09:33:02 2024	
+Fri Jul  5 09:33:02 2024	==============================================
+Fri Jul  5 09:33:02 2024	--------- Running Task
+Fri Jul  5 09:33:02 2024	==============================================
+...
+Fri Jul  5 09:33:02 2024	----------------------------------------------
+Fri Jul  5 09:33:02 2024	Phase: Running action
+Fri Jul  5 09:33:02 2024	----------------------------------------------
+Fri Jul  5 09:33:02 2024	Running command /private/var/folders/93/p7mk60kn69ggyc87dfbwtht80000gr/T/OpenJD/sample_session0bd5l1au/tmphe6bjl3c.sh
```
shidarin: I always simplify the GUIDs when giving examples to keep the line length down- so `/private/var/folders/93/p7mk60kn69ggyc87dfbwtht80000gr/` becomes something like `/private/var/folders/93/p7gr/`

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+Fri Jul  5 09:33:03 2024	Blender 4.1.1 (hash e1743a0317bc built 2024-04-16 00:06:22)
+Fri Jul  5 09:33:03 2024	Error: Cannot read file "/private/var/folders/93/p7mk60kn69ggyc87dfbwtht80000gr/T/OpenJD/sample_session0bd5l1au/3d/pavillon_barcelone_v1.2.blend": No such file or directory
+Fri Jul  5 09:33:03 2024	
+Fri Jul  5 09:33:03 2024	Blender quit
+...
+```
+
+Uh-oh, there's an error! Blender couldn't find the input file. This is because all actions in a Session are always run with their
+current working directory being the Session's temporary working directory
+(`/private/var/folders/93/p7mk60kn69ggyc87dfbwtht80000gr/T/OpenJD/sample_session0bd5l1au` in this case), the scene file argument to your
```
shidarin:
```suggestion
current working directory being the Session's temporary working directory, which was `/private/var/folders/93/p7mk60kn69ggyc87dfbwtht80000gr/T/OpenJD/sample_session0bd5l1au` in this case. In our script, we've been using a relative path...
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+```bash
+% ls /Users/myusername/blender_demo/output_frames 
+frame-001.png	frame-002.png
+```
+
+Before we continue adding functionality to your Job we're going to have you add a timeout to the Task. The timeout
+property of an action limits how long that action is allowed to run. When the time limit is reached then the
+command that's running will be canceled using the action's defined [cancelation method](2023-09-Template-Schemas#53-cancelationmethod)
+(by default, all processes in the Task's process tree are stopped). The timeout is a guard-rail to help prevent misbehaving actions
+from running forever; which will incur a monetary cost if your compute is, say, running on a cloud provider. Add a short timeout,
```
shidarin:
```suggestion
from running forever; which can incur direct costs depending on your infrastructure (such as a cloud provider).

To test the timeout functionality works as expected, add a too short timeout,
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+Fri Jul  5 09:59:28 2024	==============================================
+Fri Jul  5 09:59:28 2024	--------- Running Task
+Fri Jul  5 09:59:28 2024	==============================================
+...
+Fri Jul  5 09:59:30 2024	TIMEOUT - Runtime limit reached at 2024-07-05T14:59:30Z. Canceling action.
+Fri Jul  5 09:59:30 2024	Canceling subprocess 88373 via termination method at 2024-07-05T14:59:30Z.
+...
+```
+
+Finally, to complete this part of the guide add a Step that encodes the video to your Job. This will follow the same
```
shidarin:
```suggestion
As expected, a timeout of 2 seconds is too short to allow the Blender render to complete, and the Job is cancelled. 

Make sure to modify the timeout to a larger value before going forward.

To complete this part of the guide, add a Step that encodes the video to your Job. This will follow the same
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+                -vf "scale=in_color_matrix=bt709:out_color_matrix=bt709" \
+                -frames:v 300 -c:v libx264 -preset fast \
+                -color_range tv -colorspace bt709 -color_primaries bt709 -color_trc iec61966-2-1 \
+                -movflags faststart "$OUTPUT_FILENAME"
+
+```
+
+Notice that the EncodeVideo Step differs from the BlenderRender Step in that it includes the definition of the Step's
+[dependencies](2023-09-Template-Schemas#32-stepdependency). This ensures that when you submit the Job to a compute cluster
+to run that the `BlenderRender` Step will run successfully to completion before the `EncodeVideo` Step is run.
```
shidarin:
```suggestion
to run that the `BlenderRender` Step will run successfully to completion before the `EncodeVideo` Step is started. 
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+
+As a final note, if your shell scripts in your actual Jobs are large then you may prefer to host them on a shared network fileshare
```
shidarin:
```suggestion
If your shell scripts in your actual Jobs are large then you may prefer to host them on a shared network fileshare
```

Not really final, just the end of the part.

shidarin: I also could take or leave this final paragraph

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+instead of embedding them directly within a Job Template. If you do so, then to avoid a potential command-injection attack on your
+Jobs we remind you to ensure that no unauthorized users (including any running Job on the cluster) can possibly overwrite your scripts. 
+
+### 2.2.2. Parameterizing the template
+
+At this point you have a Job Template that has a specific scene file and frame range hard-coded into it. You can make this more
```
shidarin: Move the hard-coded bit up front to make it clear
```suggestion
At this point you have a Job Template that has hard-coded values for the scene file and frame range. You can make this more
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+Then make sure that the syntax is correct:
+
+```bash
+% openjd check job.template.yaml
+Template at 'job.template.yaml' passes validation checks.
+```
+
+Next, if you run the updated template and provide a value for this parameter then the behavior of the Job will not change because we
+are still hard-coding the location of the scene file within the template.
+
```
shidarin: I think this is a little tooooo step by step. It'd be more clear if you defined and referenced the parameter in a single step

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+% openjd run --step BlenderRender job.template.yaml -p SceneFile=/does/not/exist
+...
+Fri Jul  5 14:18:11 2024	----------------------------------------------
+Fri Jul  5 14:18:11 2024	Phase: Running action
+Fri Jul  5 14:18:11 2024	----------------------------------------------
+Fri Jul  5 14:18:11 2024	Running command /private/var/folders/93/p7mk60kn69ggyc87dfbwtht80000gr/T/OpenJD/sample_sessionvg8p3pkk/tmpend9rf4c.sh
+Fri Jul  5 14:18:11 2024	Command started as pid: 24863
+Fri Jul  5 14:18:11 2024	Output:
+Fri Jul  5 14:18:12 2024	Blender 4.1.1 (hash e1743a0317bc built 2024-04-16 00:06:22)
+Fri Jul  5 14:18:12 2024	Error: Cannot read file "/does/not/exist": No such file or directory
```
shidarin: and skip the error, just provide the value that used to be where `{{Param.SceneFile}}` is now.

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+Tasks run: 1
+```
+
+Repeating the same for the other values that we want to parameterize:
+
+1. The location of the `output_frames` directory and the output animation file's name are both [`PATH` type job parameters](2023-09-Template-Schemas#22-jobpathparameterdefinition);
+2. The starting and ending frame number for the animation are both [`INT` type job parameters](2023-09-Template-Schemas#23-jobintparameterdefinition); and
+3. The name of the job can be a [`STRING` type job parameter](2023-09-Template-Schemas#21-jobstringparameterdefinition).
+
+Pause here and try to update those yourself before continuing. The result will end up looking equivalent to:
```
shidarin:
```suggestion
The result will end up looking equivalent to:
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+          type: TEXT
+          filename: render.sh
+          runnable: true
+          data: |
+            #!/bin/bash
+
+            # Return an error code if any command in the script fails.
+            set -eou pipefail
+
+            TMPFILE=$(mktemp)
```
shidarin, line 750: Needs the comments

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+Fri Jul  5 14:28:30 2024	==============================================
+...
+Session ended successfully
+
+Job: DemoJob
+Step: BlenderRender
+Duration: 18.346753125 seconds
+Tasks run: 1
+```
+
```
shidarin, line 825: Note here that we're still running at low resolution for testing.

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+[path mapping](How-Jobs-Are-Run#path-mapping) mechanism to assist remapping the location of files directories from your workstation
+to their location on the compute host where the Job will run.
+
+The template that you've written so far is already set up to make use of the path mapping mechanism by using PATH type job parameters,
+rather than STRING types for parameters that are files and directories. The PATH type parameters will automatically have any path mapping
+rules applied to them when resolving the value in a template.
+
+You can use the CLI to test the path mapping of this job template as follows. We, ultimately, want our job to use the files in our
+current directory when running, but pretend that you're submitting from a different workstation where the files are located
+in the `/mnt/shared/demo` directory. First, create an environment variable that contains the path mapping rules
```
shidarin: This isn't super clear and could use some rewording

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+[according to the specification](How-Jobs-Are-Run#path-mapping) as though you are submitting from that fictional workstation and
+that the Job will be run on your workstation:
+
+```bash
+PATH_MAPPING_RULES="{\"version\":\"pathmapping-1.0\", \"path_mapping_rules\": [{\"source_path_format\": \"POSIX\", \"source_path\": \"/mnt/shared/demo\", \"destination_path\": \"$(pwd)\"}]}"
```
shidarin: Before dumping this on the reader, it would be best to break it out in a more readable format and describe what literal mapping this path mapping json will accomplish

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+Fri Jul  5 17:09:48 2024	Read blend: "/Users/myusername/blender_demo/3d/pavillon_barcelone_v1.2.blend"
+...
+```
+
+This works because the `pavillon_barcelone_v1.2.blend` file references the files that it needs (the textures beside it in the directory)
+with relative path references. If, on the other hand, `pavillon_barcelone_v1.2.blend` referenced files with absolute paths then this would
+not work as-is because Blender does not know where the texture files that it needs have been relocated to on the compute host. You can
+demonstrate this by opening the scene in Blender, changing the external references to absolute 
+(`File -> External Data -> Make paths absolute`), saving, and then moving the scene file and textures to a different directory and
+re-running the job. You'll notice that the resulting images are all pink due to missing textures.
```
shidarin: This feels like a complication that might not be worth mentioning at the target reader level. Maybe just a quick aside about "The Blender file only contains internal or relative references"

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+the features of your IDE, such as syntax highlighting and validation, that help you author shell scripts. The modified script to render
+a single frame at a time looks like:
+
+```bash
+#!/bin/bash
+
+# Return an error code if any command in the script fails.
+set -eou pipefail
+
+TMPFILE=$(mktemp)
```
shidarin, line 955: Copy the comments from above

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+Mon Jul  8 10:08:24 2024	
+Mon Jul  8 10:08:24 2024	==============================================
+Mon Jul  8 10:08:24 2024	--------- Running Task
+Mon Jul  8 10:08:24 2024	==============================================
+Mon Jul  8 10:08:24 2024	Parameter values:
+Mon Jul  8 10:08:24 2024	Frame(INT) = 100
+...
+```
+
+When you run the CLI without the `-tp` argument, then you'll see all of the individual Tasks in the Step being run independently:
```
shidarin: If you were to run the CLI without the argument.... but if you did, then it would try and render all 380 frames..

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
+# Introduction to Creating a Job
+
+This page guides you through developing a Job using Open Job Description. We'll explain elements of the Job Template bit by bit
```
shidarin: First paragraph should be rewritten to explicitly call out the topics we're covering below, ideally with what we expect them to be web search'd by. So make sure to mention OpenJD here, Path Mapping, Chunking, etc.

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+          range: "1-380:11"
+        # RangeEnd = 11,22,33,44,...373,380
+        - name: RangeEnd
+          type: INT
+          range: "11-380:11,380"
+      # Tasks = (1,11), (12,22), (23,33), ... etc
+      combination: "(RangeStart,RangeEnd)"
+...
+```
+
```
shidarin, line 1228: Give the example of what the above will result in, and what the `:` is doing, as this is non-obvious. Because :11 results in an integer being returned for the range every 11 values.. etc.

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+Mon Jul  8 12:23:56 2024	==============================================
+Mon Jul  8 12:23:56 2024	--------- Running Task
+Mon Jul  8 12:23:56 2024	==============================================
+Mon Jul  8 12:23:56 2024	Parameter values:
+Mon Jul  8 12:23:56 2024	RangeStart(INT) = 1
+Mon Jul  8 12:23:56 2024	RangeEnd(INT) = 11
+...
+```
+
+The complete Job Template is:
```
shidarin:
```suggestion
The Job Template is now:
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+          range:
+          - "1..40"
+          - "41..80"
+          - "81..120"
+... and so on until
+          - "361-380"
+...
+```
+
+Also remove the definitions of the `FrameStart` and `FrameEnd` Job Parameters from the template because they are no longer used,
```
shidarin: I don't think we need the rest of this example to be fully demonstrated. We can list the caveats and gotchas and move on.

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+OUTDIR="$2"
+FRAME="$3"
+
+blender --background "$SCENE" \
+  --render-output "$OUTDIR"/frame-### \
+  --render-format PNG --use-extension 1 \
+  --render-frame "${FRAME}"
+```
+
+or you need to modify the template so that it has both a testing and production mode. We'll do the later in this guide to demonstrate
```
shidarin, line 1608: I think we should break out this guide is long enough to break out into multiple pages. If we do that, feel free to ignore some of my earlier comments are removals.

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+using [Environments](2023-09-Template-Schemas#4-environment) in a Job Template. Think of Environments as encapsulating commands that you
+can run [before any Task has run, and after all Tasks have completed](How-Jobs-Are-Run#sessions) (whether successfully or not).
+Since they're just arbitrary commands of your choosing they can do anything that you would like, but one of their use cases is to
+set up or modify the environment where your Tasks will be running. For this case, modify the `render.sh` embedded file
+so that the testing code is only run if the environment variable `TESTING_TEMPLATE` has the value `true`:
```
shidarin:
```suggestion
set up or modify the environment where your Tasks will be running.

For this case, modify the `render.sh` embedded file
so that the testing code is only run if the environment variable `TESTING_TEMPLATE` has the value `true`:
```

wiki/Introduction-to-Creating-a-Job.md
```diff
@@ -0,0 +1,1818 @@
...
+bpy.context.scene.cycles.samples = 100
+EOF
+fi
+
+SCENE="$1"
+OUTDIR="$2"
+FRAME="$3"
+
+blender --background "$SCENE" \
+  --python "$TMPFILE" \
```
shidarin, line 1641: this won't error with an empty file?

shidarin (dismissed):

wiki/Introduction-to-Creating-a-Job.md
ddneilson: Nope. Turns out that an empty file is a perfectly valid python script ;-)

wiki/Introduction-to-Creating-a-Job.md
ddneilson: command-line args. Clarified in doc

wiki/Introduction-to-Creating-a-Job.md
ddneilson: I've adjusted this significantly. Let me know what you think.

wiki/Introduction-to-Creating-a-Job.md
ddneilson: Added something.

wiki/Introduction-to-Creating-a-Job.md
ddneilson: Thanks. Condensed.

wiki/Introduction-to-Creating-a-Job.md
ddneilson: Reworded.

wiki/Introduction-to-Creating-a-Job.md
ddneilson: Fixed up with the rewording above. Let me know what you think?

wiki/Introduction-to-Creating-a-Job.md
ddneilson: Yeah, it's a bit much.

wiki/Introduction-to-Creating-a-Job.md
ddneilson: Reworded this a bit.

wiki/Introduction-to-Creating-a-Job.md
ddneilson: I've added an aside to help.

shidarin (approved):

Issue #53, opened 2024-12-02
RFC: Task chunking - running multiple tasks with one command


mwiebe:
* **Pull Request**: https://github.com/OpenJobDescription/openjd-specifications/pull/54
* **Discussion Thread(s)**: 
    * https://github.com/OpenJobDescription/openjd-specifications/discussions/52

## Description

Chunking frames together is a commonly supported solution for amortizing application and scene loading overhead, and Open Job Description should support it.  This RFC adds chunking to job templates with the goal that it be easy to author and read templates that work well with both contiguous frame ranges and pick up renders, and is not to much work to implement.

## Roles

| Role | User
| ---- | ----
| Proposed By | @mwiebe 
| Author(s)   | @mwiebe 

## Workflow

- [x] Tracking issue created (label: `rfc/proposed`)
- [x] RFC pull request submitted and ready for discussion (label: `rfc/exploring`)
- [x] Last call for comments (labels: `rfc/exploring` and `rfc/final-comments`)
- [x] Accepted and merged RFC pull request (label: `rfc/accepted-future`)
- [x] Green-light for inclusion in a draft specification, and the author is creating and iterating on pull requests (label: `rfc/accepted-draft`)
- [x] Pull requests are merged in to a draft specification (label: `rfc/accepted-staged`)
- [x] Officially published in a non-draft revision of the specification (label: `rfc/released`)

Please close this tracking issue when the proposal enters the `Released` stage of the process.

## Open Points

For easier discovery, especially if there is a lot of discussion on this issue, then please keep this section updated
with brief summaries and pointers to the main points of discussion.

---

> The author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue.

crowecawcaw:
RFC was accepted.

PR #54, opened 2024-12-02
feat(rfc): Write RFC 0001 to propose task chunking support

mwiebe:
## Request for comment on task chunking proposal

**Tracking Issue**: https://github.com/OpenJobDescription/openjd-specifications/issues/53

Chunking frames together is a commonly supported solution for amortizing application and scene loading overhead, and Open Job Description should support it. This RFC adds chunking to job templates with the goal that it be easy to author and read templates that work well with both contiguous frame ranges and pick up renders, and is not to much work to implement.

Here's a link directly to the rendered RFC: https://github.com/mwiebe/openjd-specifications/blob/task-chunking-rfc/rfcs/0001-task-chunking.md

---

### DIFF
--- a/rfcs/0001-task-chunking.md
+++ b/rfcs/0001-task-chunking.md
@@ -1,0 +1,452 @@
+
+* Feature Name: Task Chunking
+* RFC Tracking Issue: https://github.com/OpenJobDescription/openjd-specifications/issues/53
+* Start Date: 2024-11-27
+* Specification Version: 2023-09 extension TASK_CHUNKING
+* Accepted On: in draft
+* Depends On: RFC 0002 (https://github.com/OpenJobDescription/openjd-specifications/issues/57)
+
+## Summary
+
+This RFC proposes to address application and scene loading overhead by adding a mechanism
+for chunking to the Open Job Description job template specification. Chunking is a way to
+improve resource utilization in exchange for larger task runtimes. Render jobs often have
+a single scene file that defines a scene graph including an animated camera, and output
+one frame of a shot for each task. Instead of rendering one frame per CLI command,
+with chunking the scheduler dispatches a set of frames. The chunk may be a contiguous range
+like frames "1-10", or a selection of them like "1,3,4,7,8,10".
+
+This proposal complements task stickiness, the approach already supported for loading
+the application as a background daemon and then scheduling individual tasks to it without
+closing the application.
+
+## Basic Examples
+
+### Contiguous chunks
+
+This sample shows a job for a `render` CLI tool that supports `-start` and `-end` frame range parameters.
+With the default `Frames` of "1,10-12,18-50" and `ChunkSize` of 10, a likely
+division of the input frames into contiguous chunks is "1-1", "10-12", "18-27",
+"28-37", "38-47", "48-50".
+
+New syntax is annotated by comments:
+
+```yaml
+specificationVersion: 'jobtemplate-2023-09'
+extensions:
+- TASK_CHUNKING
+name: Contiguous Chunks
+parameterDefinitions:
+  - name: SceneFile
+    type: PATH
+    objectType: FILE
+    dataFlow: IN
+  - name: Frames
+    type: STRING
+    default: 1,10-12,18-50
+  - name: ChunkSize
+    type: INT
+    default: 10
+steps:
+  - name: Render
+    parameterSpace:
+      taskParameterDefinitions:
+        - name: Frame
+          ## The CHUNK[INT] task parameter type is added by this proposal
+          type: CHUNK[INT]
+          range: "{{Param.Frames}}"
+          ## The chunks field must be provided for type CHUNK[INT]
+          chunks:
+            ## Required, is how big to make each chunk by default
+            defaultTaskCount: "{{Param.ChunkSize}}"
+            ## Optional and can be ignored by the scheduler. If supported,
+            ## the scheduler adjusts the task count for chunks to match
+            ## this runtime. The value is in seconds, so 900 is 15 minutes.
+            targetRuntimeSeconds: 900
+            ## Required, affects the script handling the onRun action.
+            ## CONTIGUOUS means {{Task.Param.Frame}} always expands to
+            ## "<startframe>-<endframe>", e.g. "1-1" or "11-20".
+            rangeConstraint: CONTIGUOUS
+    script:
+      actions:
+        onRun:
+          command: bash
+          args: ["{{Task.File.RenderChunk}}"]
+      embeddedFiles:
+        - name: RenderChunk
+          filename: render.sh
+          type: TEXT
+          ## Because a CONTIGUOUS chunk is always like "<startframe>-<endframe>",
+          ## this code can use the cut command to get the start and end values.
+          data: |
+            START_FRAME="$(echo '{{Task.Param.Frame}}' | cut -d- -f1)"
+            END_FRAME="$(echo '{{Task.Param.Frame}}' | cut -d- -f2)"
+            render -scenefile '{{Param.SceneFile}}' \
+                -start "$START_FRAME" \
+                -end "$END_FRAME"
+```
+
+### Non-contiguous chunks
+
+This sample shows a job for the V-Ray CLI that supports arbitrary frame lists, as documented in
+[V-Ray Standalone Command Line Options](https://docs.chaos.com/display/VNS/V-Ray+Standalone+Command+Line+Options).
+Because V-Ray has a different frame range syntax than Open Job Description, the job performs
+a character substitution for the value.
+
+```yaml
+specificationVersion: 'jobtemplate-2023-09'
+extensions:
+- TASK_CHUNKING
+name: V-Ray with Non-contiguous Chunks
+parameterDefinitions:
+  - name: VRSceneFile
+    type: PATH
+    objectType: FILE
+    dataFlow: IN
+  - name: Frames
+    type: STRING
+    default: "1-60:2"
+steps:
+  - name: Render
+    parameterSpace:
+      taskParameterDefinitions:
+        - name: Frame
+          type: CHUNK[INT]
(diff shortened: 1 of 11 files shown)


rfcs/0001-task-chunking.md
```diff
@@ -0,0 +1,423 @@
...
+          type: CHUNK[INT]
+          range: "{{Param.Frames}}"
+          ## The chunks field must be provided for type CHUNK[INT]
+          chunks:
+            ## Required, is how big to make each chunk by default
+            defaultTaskCount: "{{Param.ChunkSize}}"
+            ## Optional and can be ignored by the scheduler. If supported,
+            ## the scheduler adjusts the task count for chunks to match
+            ## this runtime. The value is in seconds, so 900 is 15 minutes.
+            targetRuntime: 900
```
prestomation: It strikes me that we have to tell the user the units are 'seconds' in the comment. It would be more self-explanatory for template writers what the unit was if it was contained in the key name, like `targetRuntimeSeconds`.   That said the existing `timeout` is in seconds so probably better to be consistent here.  Big brain thinking would be supporting ISO8601 durating strings but almost certainly overkill.

All that said, thinking backwards to customers, this seems like a parameter that might be set by an integrated submitter using heuristics of scene size, local render time, whatever so probably not a huge need to be human readable.

mwiebe: It's a good point, maybe we should consider adding the unit? The human readability of the template itself is one of [the tenets](https://github.com/OpenJobDescription/openjd-specifications/wiki/Design-Tenets):

>  Human readable and writable - The artifacts that users create to interact with Open Job Description are easily understood and authored by a human with nothing more than a text editor. They lead to behavior that is easy to understand from a plain reading of the artifact, and can be reproduced in any compatible compute environment.

We could consider migrating the timeouts as well, introducing `timeoutSeconds` as a mutually exclusive synonym for backwards compatibility.

Ahuge:
Love this RFC, thanks Mark.

~~Question for you, do we have a reference to the rangeConstraint value to determine if the `{{Task.Param.Frame}}` value is a NONCONTIGUOUS or a CONTIGUOUS value?~~

~~Perhaps that isn't something we need to worry about, I am just getting my head around this RFC but I wanted to ask if that had been considered.~~

****Edit** the `rangeConstraint` determines how OpenJD will pass chunks to the task, that makes sense.
The `CONTIGUOUS` vs `NONCONTIGUOUS` value refers to if the task/adaptor supports `NONCONTIGUOUS` ranges.

I was implementing a POC for the NukeAdaptor and by default it currently only supports `(\d+)-(\d+)` frame ranges (`CONTIGUOUS`) but because we're running an adaptor workflow, we could add `NONCONTIGUOUS` functionality


Please correct me if I am wrong in that assumption.


**Perhaps having a third example would help. Where the frame range passed to the job parameter is a complex/NONCONTIGUOUS range but setting the `rangeConstraint` value to be `CONTIGUOUS` shows that the task only ever receives `CONTIGUOUS` chunks

mwiebe:
> Love this RFC, thanks Mark.

Awesome, thanks!

> Question for you, do we have a reference to the rangeConstraint value to determine if the `{{Task.Param.Frame}}` value is a NONCONTIGUOUS or a CONTIGUOUS value?

I expect that the person setting the rangeConstraint and the person writing the step script are the same, so for a particular step it would always be one or the other. Do you have an idea about when you might reach for that value programmatically from the step script, versus just implementing one or the other behavior directly?

> Perhaps that isn't something we need to worry about, I am just getting my head around this RFC but I wanted to ask if that had been considered.

I had considered a slightly different angle, of whether it makes sense for someone submitting a job to control this option. My conclusion was it's better not to, and the "`@fmtstring` selection" Design Rationale section talks about it.

mwiebe:
> I was implementing a POC for the NukeAdaptor and by default it currently only supports `(\d+)-(\d+)` frame ranges (`CONTIGUOUS`) but because we're running an adaptor workflow, we could add `NONCONTIGUOUS` functionality
> 
> Please correct me if I am wrong in that assumption

Yes this is exactly the idea. Then when someone submits a pickup render with a few scattered frames, it can run with chunks like "7,12,25".

mwiebe:
> **Perhaps having a third example would help. Where the frame range passed to the job parameter is a complex/NONCONTIGUOUS range but setting the rangeConstraint value to be CONTIGUOUS shows that the task only ever receives CONTIGUOUS chunks

This is a good idea, maybe I can do that in the first example by setting a non-contiguous default there with a comment explaining how it will get chunked.

UPDATE: I've changes the first example's default frames to something noncontiguous and added a description of likely chunk selection. Thanks for the suggestion!

rfcs/0001-task-chunking.md
mwiebe: I've changed the name to `targetRuntimeSeconds` and added a design rationale explaining the reasoning to break with precedent.

rfcs/0001-task-chunking.md
```diff
@@ -0,0 +1,440 @@
...
++ ```
++
++ See section `<IntTaskParameterDefinition>` for the definitions of `<IntRangeExpr>` and `<intstring>`.
++
++ 1. *name* — The name of the parameter.
++ 2. *type* — The literal "CHUNK[INT]", defining this parameter as integer valued and processed
++     as chunks.
++ 3. *range* — The list of values that the parameter takes on to define Tasks of the Step.
++ 4. *chunks* — Specifies how to form sets of values into chunks.
++     1. *defaultTaskCount* — How many tasks to combine into a single chunk by default.
```
jairaws, line 261: Why are we using the `default` prefix? Is it because this value can be overridden at runtime based on `targetRuntimeSeconds` or by the remaining frames in a range? If so, is this the same as `maxTaskCount`?

jairaws (approved):

rfcs/0001-task-chunking.md
mwiebe: Thanks for reviewing the proposal! It's called default to indicate the scheduler should use this value if it doesn't have a better value to use. It's not the same as `maxTaskCount` because the scheduler can choose larger values, like when chunks are finishing quicker than in `targetRuntimeSeconds`.

rfcs/0001-task-chunking.md
```diff
@@ -0,0 +1,452 @@
...
++  A `<ChunkIntTaskParameterDefinition>` is the object:
++
++ ```yaml
++   name: <Identifier>
++   type: "CHUNK[INT]"
++   range: <IntRangeList> | <IntRangeExpr>
++   chunks:
++     defaultTaskCount: <integer> | <intstring> # @fmtstring
++     targetRuntimeSeconds: <integer> | <intstring> # @optional @fmtstring
++     rangeConstraint: "CONTIGUOUS" | "NONCONTIGUOUS"
```
crowecawcaw, line 251: Would it make sense to make rangeContraint optional and have a default value of NONCONTIGUOUS? Pedantically, the NONCONTIGUOUS option is a lack of a constraint. I'd guess that NONCONTIGUOUS would be the much more common option, so it'd simplify the parameter definition for many templates.

crowecawcaw (approved):

rfcs/0001-task-chunking.md
mwiebe: I made it required so that templates will be more understandable without having to know what the default is. I would have picked `CONTIGUOUS` as the default, and `NONCONTIGUOUS` as the more general option you can opt into. I was thinking of CLI commands that have start/end frame options.

rfcs/0001-task-chunking.md
```diff
@@ -0,0 +1,452 @@
...
+or in chunks of 50 frames. See
+[this pull request description](https://github.com/aws-deadline/deadline-cloud-for-after-effects/pull/77)
+for details.
+
+Our model of chunk runtime starts an application and loads a scene, then runs tasks in parallel
+until all the tasks in the chunk are done.
+
+![chunk example runtime](.images/0001-task-chunking-chunk-example-runtime.svg)
+
+Here's an example timeline for running 15 tasks with this chunking on one worker host,
```
eherozhao, line 193: I am not sure if I misunderstand this.
Does these 15 tasks mean 15 frames here? or openjd Tasks? I am thinking that we are talking about 3 openjd tasks with 6 frames chunk size but I am a bit confused about the terminalogy here.

rfcs/0001-task-chunking.md
ryyakobe: Maybe we could consider renaming it to `initialTaskCount` or even `initialChunkSize` to emphasize that it's the starting point for chunk sizing and avoid confusion with optional parameters that have predefined defaults?

rfcs/0001-task-chunking.md
```diff
@@ -0,0 +1,452 @@
...
++ individual Tasks. A Chunk is a set of Tasks with all of the Task parameter values identical except for the chunked
++ Task parameter. It takes values from an integer range expression like "1-3" or 1-3,5,7" depending on whether
++ the chunks are constrained to be contiguous or not.
+
+  All failures and cancellations in a **Session** are terminal for the **Session**, as the system generally does not know
+```
+
+> Modifications to the [Introduction to Creating a Job](https://github.com/OpenJobDescription/openjd-specifications/wiki/Introduction-to-Creating-a-Job) wiki page.
+
+Modify the section [2.2.4.2. Multiple Frames Per Task](https://github.com/OpenJobDescription/openjd-specifications/wiki/Introduction-to-Creating-a-Job#2242-multiple-frames-per-task)
```
ryyakobe, line 345: Is this a TODO item?

ryyakobe (approved):

rfcs/0001-task-chunking.md
mwiebe: Yes, task == frame here, so it means 15 tasks and 15 frames. A chunk of 6 tasks is then also a chunk of 6 frames.

PR #65, opened 2025-01-30
feat: Merge RFC 0001 into the 2023-09 template schema wiki

mwiebe:
## PR to merge RFC 0001 into specification wiki

This PR merges the content from https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md into the 2023-09 template schema wiki.

Content not specifically copied from the RFC:
* Added the `@extension EXTENSION_NAME` annotation to indicate which parts of the specification come from extensions.
* Added starting documentation about using the extension in the "Introduction to Creating a Job" wiki. This will be updated further once the extension is implemented in the CLI command and libraries.
* Updated the accepted dates of the two RFCs.

---

### DIFF
--- a/wiki/2023-09-Template-Schemas.md
+++ b/wiki/2023-09-Template-Schemas.md
@@ -22,6 +22,8 @@ Notations used in this document to annotate aspects of the schema definition:
    * `@fmtstring[host]` - The value is evaluated at runtime on the worker host on which the job is running. The value is otherwise
      evaluated when the job template is submitted and the render manager is constructing a job.
 * `@optional` - The annotated property is optional.
+* `@extension EXTENSION_NAME` - The annotated property is available only when extension EXTENSION_NAME is requested
+  in the `extensions` list of the template. Implementations can choose which extensions to support.
 
 ## 1. Root elements
 
@@ -645,8 +647,11 @@ Where:
 Definition of a single task parameter; its name, type, and the range of values that it takes.
 
 ```bnf
-<TaskParameterDefinition> ::= <IntTaskParameterDefinition> | <FloatTaskParameterDefinition> |
-                              <StringTaskParameterDefinition> | <PathTaskParameterDefinition>
+<TaskParameterDefinition> ::= <IntTaskParameterDefinition> |
+                              <FloatTaskParameterDefinition> |
+                              <StringTaskParameterDefinition> |
+                              <PathTaskParameterDefinition> |
+                              <ChunkIntTaskParameterDefinition> # @extension TASK_CHUNKING
 ```
 
 ##### 3.4.1.1. `<IntTaskParameterDefinition>`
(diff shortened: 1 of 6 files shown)


wiki/2023-09-Template-Schemas.md
```diff
@@ -801,6 +806,46 @@ using the following names:
...
+A `<ChunkIntTaskParameterDefinition>` is the object:
+
+```yaml
+  name: <Identifier>
+  type: "CHUNK[INT]"
+  range: <IntRangeList> | <IntRangeExpr>
+  chunks:
+    defaultTaskCount: <integer> | <intstring> # @fmtstring
+    targetRuntimeSeconds: <integer> | <intstring> # @optional @fmtstring
+    rangeConstraint: "CONTIGUOUS" | "NONCONTIGUOUS"
```
epmog, line 821: I'll double check the RFC, but this is distinctly not a fmtstring? So users can't specify this with a job parameter and it must be hard-coded?

wiki/How-Jobs-Are-Run.md
```diff
@@ -4,33 +4,37 @@ Before reading this we recommend diving into [How Jobs Are Constructed](How-Jobs
...
-what state a failure or cancellation leaves the **Session** in. Think of a Task failure which inadvertently terminates 
-the container that was set up in an **Environment** for the Tasks to run 
-within. A failed Task will result in a failed **Session**, which will subsequently exit all **Environments** that the 
+Once the **Environments** have been created and entered for a **Session**, a series of Tasks are run within that
+**Session** and the **Environments**. Tasks from any Step in a Job can run within the same **Session** provided that
+the set of **Environments** that are required to run those Tasks are identical. If the parameter space of a Step
+includes a task parameter with type `CHUNK[INT]`, then the **Session** runs Chunks for that Step instead of
+individual Tasks. A Chunk is a set of Tasks with all of the Task parameter values identical except for the chunked
+Task parameter. It takes values from an integer range expression like "1-3" or 1-3,5,7" depending on whether
+the chunks are constrained to be contiguous or not.
```
epmog: Food for thought, I wonder if we should be highlighting in these other documents when functionality is part of an extension. I do see it like that in "Creating A Job".

A user comes here to read this and thinks chunks are always available to be running jobs, when in actuality, it depends on the scheduler supporting that extension and the user specifying it.

wiki/2023-09-Template-Schemas.md
mwiebe: Correct, I wrote a rationale for that here: https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md#selection-of-fmtstring-fields

wiki/How-Jobs-Are-Run.md
mwiebe: I've added a parenthetical note here to indicate that.

epmog (approved):

Issue #85, opened 2025-04-09
RFC: openjd_redacted_env: setting environment variables containing sensitive info which should be redacted from logs


baxeaz:
* **Pull Request**: https://github.com/OpenJobDescription/openjd-specifications/pull/86
* **Original Issue**: https://github.com/OpenJobDescription/openjd-specifications/issues/62

## Description

Setting environment variables in an openjd Environment such that they are persisted in subsequent actions run while the Environment is still active currently requires [echoing them to stdout with the openjd_env: key=value format](https://github.com/OpenJobDescription/openjd-specifications/blob/711dab30382579d060863581b56f5266c16430da/wiki/2023-09-Template-Schemas.md#4-environment) which exposes potentially sensitive information such as credentials to logs.

This RFC proposes a new token, openjd_redacted_env, which sets environment variables exactly like openjd_env but defaults to redacting the potentially sensitive information from logs.

## Roles

| Role | User
| ---- | ----
| Proposed By | @baxeaz 
| Author(s)   | @baxeaz 

## Workflow

- [x] Tracking issue created (label: `rfc/proposed`)
- [x] RFC pull request submitted and ready for discussion (label: `rfc/exploring`)
- [x] Last call for comments (labels: `rfc/exploring` and `rfc/final-comments`)
- [x] Accepted and merged RFC pull request (label: `rfc/accepted-future`)
- [ ] Green-light for inclusion in a draft specification, and the author is creating and iterating on pull requests (label: `rfc/accepted-draft`)
- [ ] Pull requests are merged in to a draft specification (label: `rfc/accepted-staged`)
- [ ] Officially published in a non-draft revision of the specification (label: `rfc/released`)

Please close this tracking issue when the proposal enters the `Released` stage of the process.

## Open Points

For easier discovery, especially if there is a lot of discussion on this issue, then please keep this section updated
with brief summaries and pointers to the main points of discussion.

---

> The author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue.

crowecawcaw:
As a design inspiration, GitHub automatically redacts strings that are stored as secrets from its logs: https://docs.github.com/en/actions/security-for-github-actions/security-guides/about-secrets#using-your-secrets-in-workflows

As an idea, we could also specify in the OJD spec that implementations should continue to redact the redacted env value from the rest of the logs in the lifetime of their environment.

baxeaz:
[...]

PR #86, opened 2025-04-09
feat(rfc): Write RFC 0003 to propose a method for redacting sensitive…

baxeaz:
… environment variable values written to stdout.

## PR for RFC

**Tracking Issue**: https://github.com/OpenJobDescription/openjd-specifications/issues/85

This is a request for comments about redacted environment variables through openjd_redacted_env

Rendered RFC: [here](https://github.com/baxeaz/openjd-specifications/blob/mainline/rfcs/0003-redacted-env-vars.md)

### DIFF
--- a/rfcs/0003-redacted-env-vars.md
+++ b/rfcs/0003-redacted-env-vars.md
@@ -1,0 +1,225 @@
+- Feature Name: Redacted Environment Variables
+- RFC Tracking Issue: https://github.com/OpenJobDescription/openjd-specifications/issues/85
+- Start Date: 2025-04-09
+- Specification Version: 2023-09 extension REDACTED_ENV_VARS
+- Accepted On: (fill in with the date that the proposal was accepted: YYYY-MM-DD)
+
+## Summary
+
+Setting environment variables in an openjd Environment such that they are persisted in subsequent actions run while the Environment is still
+active currently requires [echoing them to stdout with the openjd_env: key=value format](https://github.com/OpenJobDescription/openjd-specifications/blob/711dab30382579d060863581b56f5266c16430da/wiki/2023-09-Template-Schemas.md#4-environment)
+which exposes potentially sensitive information such as credentials to logs. This RFC proposes a new token, openjd_redacted_env, which sets
+environment variables exactly like openjd_env but defaults to redacting the potentially sensitive information from logs for the initial and
+any future occurrences.
+
+The feature is only used to redact the environment variable in log output. Environment variables for a process can be observed by other
+processes on the same machine. Secure handling of secrets is still the responsibility of those authoring job templates.
+
+## Basic Examples
+
+This example shows a very basic (somewhat contrived) example where a job environment is responsible for initially retrieving and setting
+credentials in an environment variable and then a step is run which consumes the credentials.
+
+```yaml
+specificationVersion: "jobtemplate-2023-09"
+extensions:
+  - REDACTED_ENV_VARS
+name: Env Enter Setting Vars
+description: |
+  Setting environment variables in a way which won't expose secret values to logs
+
+jobEnvironments:
+  - name: JobEnvVars
+    variables:
+      PYTHONUNBUFFERED: "True"
+    script:
+      actions:
+        onEnter:
+          command: python
+          args: ["{{Env.File.Enter}}"]
+        onExit:
+          command: python
+          args: ["{{Env.File.Exit}}"]
+      embeddedFiles:
+        - name: Enter
+          type: TEXT
+          data: |
+            print("Entering environment and setting var..")
+            print(f"openjd_redacted_env: SECRETVAR=SECRETVAL")
+        - name: Exit
+          type: TEXT
+          data: |
+            import os
+            # SECRETVAR is SECRETVAL here
+            print(f"SECRETVAR is {os.environ.get('SECRETVAR')}")
+steps:
+  - name: ProcessWithPython
+    script:
+      actions:
+        onRun:
+          command: python
+          args: ["{{Task.File.Run}}"]
+      embeddedFiles:
+        - name: Run
+          type: TEXT
+          data: |
+            import os
+
+            # SECRETVAR is SECRETVAL here
+            print(f"SECRETVAR is {os.environ.get('SECRETVAR')}")
+            # Run Application Consuming SECRETVAR..
+```
+
+Running this template with the currently proposed changes produces the following output.
+
+```
+0:00:00.000047  Open Job Description CLI: Session start 2025-04-11T21:50:33.781398+00:00
+0:00:00.000109  Open Job Description CLI: Running job 'Env Enter Setting Vars'
+0:00:00.000180
+0:00:00.000218  ==============================================
+0:00:00.000253  --------- Entering Environment: JobEnvVars
+0:00:00.000285  ==============================================
+0:00:00.000336  Setting: PYTHONUNBUFFERED=True
+0:00:00.000647  ----------------------------------------------
+0:00:00.000722  Phase: Setup
+0:00:00.000761  ----------------------------------------------
+0:00:00.000803  Writing embedded files for Environment to disk.
+0:00:00.001017  Mapping: Env.File.Enter -> /tmp/OpenJD/CLI-sessionj6z1b_nc/embedded_files1d4q40xw/tmpfdql519m
+0:00:00.001072  Mapping: Env.File.Exit -> /tmp/OpenJD/CLI-sessionj6z1b_nc/embedded_files1d4q40xw/tmpuhsmb_sg
+0:00:00.001198  Wrote: Enter -> /tmp/OpenJD/CLI-sessionj6z1b_nc/embedded_files1d4q40xw/tmpfdql519m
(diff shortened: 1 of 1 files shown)


rfcs/0003-redacted-env-vars.md
```diff
@@ -0,0 +1,121 @@
...
+    - name: Run
+      type: TEXT
+      data: |
+        import os
+
+        # SECRETVAR is SECRETVAL here
+        print(f"SECRETVAR is {os.environ.get('SECRETVAR')}")
+        # Run Application Consuming SECRETVAR..
+```
+
```
mwiebe, line 72: Because the log output is a major part of this RFC, it would be great to show the expected log output of the sample job. The best way I can think of to generate that is to save the template with `openjd_redacted_env` changed into `openjd_env` throughout, run `openjd run sample_template.yaml` to generate the current output, then hand-edit the output to say `openjd_redacted_env` and have redaction in it.

rfcs/0003-redacted-env-vars.md
```diff
@@ -0,0 +1,121 @@
...
+## Motivation
+
+Customers currently have a method of setting environments in a way that persists in actions run while the environment is still active, however it requires exposing those environment variable names and values to session logs.  These variables will occasionally contain sensitive information which customers may not wish to be exposed to their logs and visible with anyone who has access to those logs by default.  Workarounds such as applying data protection policies exist, but customers have [requested simpler methods](https://github.com/OpenJobDescription/openjd-specifications/issues/62) of optionally hiding these values.
+
+## Specification
+
+```
```
mwiebe: You can use ```diff here to get some syntax highlighting.

rfcs/0003-redacted-env-vars.md
```diff
@@ -0,0 +1,121 @@
...
+diff --git a/wiki/2023-09-Template-Schemas.md b/wiki/2023-09-Template-Schemas.md
+index 7307991..bcad430 100644
+--- a/wiki/2023-09-Template-Schemas.md
++++ b/wiki/2023-09-Template-Schemas.md
+@@ -989,7 +989,11 @@ Implementations of this specfication must watch STDOUT when running the `onEnter
+ 1. The regular expression `^openjd_env: (.*)$`. The captured value must be of the form `<varname>=<value>` where
+    `varname` is the name of an environment variable, and `value` is the value to assign to it. The defined value of the
+    given variable will be set for all actions that are run with the environment active.
+-2. The regular expression `^openjd_unset_env: (.*)$`. The captured value must be of the form `<varname>` where
++2. The regular expression `^openjd_redacted_env: (.*)$`. The captured value must be of the form `<varname>=<value>` where
```
mwiebe: Instead of duplicating the same text, maybe say "The captured value must be interpreted identically to `openjd_env`."

rfcs/0003-redacted-env-vars.md
baxeaz: Good call!  I actually ended up doing a sample implementation today which exposed a couple other edge cases/issues.  Will add the sample output along with updates for the items I uncovered

rfcs/0003-redacted-env-vars.md
baxeaz: Will update!

rfcs/0003-redacted-env-vars.md
baxeaz: Makes sense, will update!

rfcs/0003-redacted-env-vars.md
```diff
@@ -0,0 +1,121 @@
...
+--- a/wiki/How-Jobs-Are-Run.md
++++ b/wiki/How-Jobs-Are-Run.md
+@@ -84,6 +84,10 @@ messages to convey information about the **Action** to the render management sys
+ * `openjd_env: <var>=<value>` where `<var>` is the string name of an environment variable, and `<value>` is a string. This
+   can only be emitted by the **Action** for entering an **Environment**. It defines the value of an environment variable for
+   all subsequent **Action**s in the **Session** until the defining **Environment** is exited.
++* `openjd_redacted_env: <var>=<value>` where `<var>` is the string name of an environment variable, and `<value>` is a string. This
++  can only be emitted by the **Action** for entering an **Environment**. It defines the value of an environment variable for
++  all subsequent **Action**s in the **Session** until the defining **Environment** is exited.  `<value>` shall be redacted by the
++  application running the action in the emitted stdout line.
```
crowecawcaw: Does the redaction happen in logs for all actions within the environment or just in this particular line or onEnter action? I think it'd be better if it were redacted everywhere for some extra safety in case someone accidentally logs the env variable while running.

baxeaz: Currently I was imagining it would be just the specific line since it's a bit hard to know intent otherwise, but I think what you're saying might make more sense. The reasons not to redact elsewhere would be things like "What if it's an unrelated string that somehow collides" which might either unnecessarily redact a section of logs, or at worst give a pretty strong hint at what the redacted string was, or "What if I was intentionally trying to print it in order to debug or something" - in which case they could just switch to openjd_env.

mwiebe: Redacting later in the logs sounds like a good idea to me. Especially for debug code like "Print all the environment variables to see what's going on."

baxeaz: Adopted, added to the specification, added to my POC, updated the sample above with the new behavior.  Thanks for the suggestion!

rfcs/0003-redacted-env-vars.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Behavior when REDACTED_ENV_VARS supported but not specified
+
+In order to maintain as high a security bar as possible, this document proposes that the redaction behavior of openjd_redacted_env be honored in supporting applications even when the extension is not specified, though the environment variable would not be set.  This is a deviation from standard extension behavior where without the extension none of the custom behavior would be followed.
+
+## Prior Art
```
mwiebe, line 213: The reference to github secrets in https://github.com/OpenJobDescription/openjd-specifications/issues/85#issuecomment-2791076795 seems worth mentioning here.

baxeaz: Added, thanks!  (Wasn't quite sure which section you meant but seemed like it belonged in Prior Art to me)

crowecawcaw (dismissed):

crowecawcaw (dismissed):

baxeaz:
Please note #86  is now in final-comments - if there are any other concerns please let me know!

rfcs/0003-redacted-env-vars.md
```diff
@@ -0,0 +1,204 @@
...
+```diff
+diff --git a/wiki/2023-09-Template-Schemas.md b/wiki/2023-09-Template-Schemas.md
+--- a/wiki/2023-09-Template-Schemas.md
++++ b/wiki/2023-09-Template-Schemas.md
+@@ -989,7 +989,11 @@ Implementations of this specfication must watch STDOUT when running the `onEnter
+ 1. The regular expression `^openjd_env: (.*)$`. The captured value must be of the form `<varname>=<value>` where
+    `varname` is the name of an environment variable, and `value` is the value to assign to it. The defined value of the
+    given variable will be set for all actions that are run with the environment active.
+-2. The regular expression `^openjd_unset_env: (.*)$`. The captured value must be of the form `<varname>` where
++2. The regular expression `^openjd_redacted_env: (.*)$`. The captured value must be interpreted identically to openjd_env, and `<value>` shall be redacted with a fixed length string for this and any future occurrences from the emitted STDOUT message.  Requires the REDACTED_ENV_VARS extension, however supporting applications must honor the redaction even when the extension is not specified.
```
mwiebe: Could you break up these long lines? To see the full text in rendered view, you have to horizontally scroll the diff text block.

rfcs/0003-redacted-env-vars.md
```diff
@@ -0,0 +1,204 @@
...
+### Behavior when REDACTED_ENV_VARS supported but not specified
+
+In order to maintain as high a security bar as possible, this document proposes that the redaction behavior of openjd_redacted_env be honored in supporting applications even when the extension is not specified, though the environment variable would not be set.  This is a deviation from standard extension behavior where without the extension none of the custom behavior would be followed.
+
+## Prior Art
+
+This feature is intended to provide nearly identical behavior to `openjd_env` except with protection of redacting `<value>` to give
+customers a drop in replacement for cases handling sensitive environment variables.
+
+The goal of this proposal is to protect secrets in a manner similar to [GitHub Secrets](https://docs.github.com/en/actions/security-for-github-actions/security-guides/about-secrets#using-your-secrets-in-workflows) but with the added protection of continuing to redact them in the session logs once declared as redacted through openjd_redacted_env.
```
mwiebe: Not sure if "added protection" is right? The github secrets description sounds like it does the same thing.

rfcs/0003-redacted-env-vars.md
baxeaz: Fixed!

rfcs/0003-redacted-env-vars.md
baxeaz: Hmm Stephen also pointed that out previously, thought I'd fixed it, thanks!

mwiebe (approved):

Issue #92, opened 2025-10-02
RFC: FEATURE_BUNDLE_1: Collection of Small Improvements


edwards-aws:
* **Pull Request**: (Add the URL to the pull request here.)
* **Discussion Thread(s)**:
    * https://github.com/OpenJobDescription/openjd-specifications/discussions/77

## Description

There are many small improvements that could be made to Open Job Description that would make it easier and more expressive to write job templates. It is proposed that many of these could be bundled into an extension called FEATURE_BUNDLE_1.

These features would include:

1. Name length limit increases for parameter names, job names, step names, environment names, file names, and identifiers.
1. Job Parameter count limit increases
1. Format string enhancements for timeout
1. EOL control in embedded files
1. Syntax sugar for script interpreters commonly used in templates in `<Action>` objects. This would include bash, python, cmd, powershell, and node.
1. `onRun` in `<StepActions>` action property generalization

## Roles

| Role | User
| ---- | ----
| Proposed By | @mwiebe @epmog 
| Author(s)   | @edwards-aws 

## Workflow

- [x] Tracking issue created (label: `rfc/proposed`)
- [x] RFC pull request submitted and ready for discussion (label: `rfc/exploring`)
- [x] Last call for comments (labels: `rfc/exploring` and `rfc/final-comments`)
- [x] Accepted and merged RFC pull request (label: `rfc/accepted-future`)
- [ ] Green-light for inclusion in a draft specification, and the author is creating and iterating on pull requests (label: `rfc/accepted-draft`)
- [ ] Pull requests are merged in to a draft specification (label: `rfc/accepted-staged`)
- [ ] Officially published in a non-draft revision of the specification (label: `rfc/released`)

Please close this tracking issue when the proposal enters the `Released` stage of the process.

## Open Points

For easier discovery, especially if there is a lot of discussion on this issue, then please keep this section updated
with brief summaries and pointers to the main points of discussion.

---

> The author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue.

edwards-aws:
Please note https://github.com/OpenJobDescription/openjd-specifications/pull/95 is now in final-comments - if there are any other concerns please let me know!

PR #95, opened 2025-10-24
feat(rfc): Write RFC 004 to propose a grab bag of extensions

edwards-aws:
**Tracking Issue**: https://github.com/OpenJobDescription/openjd-specifications/issues/92

This is a request for comments about Collection of Small Improvements.

---


rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
+* Improved control over embedded file line endings
+* Syntax sugar for common script interpreters in `<Action>` objects
+
+## Basic Examples
+
+### Increased Name Length Limits
+
+```yaml
+specificationVersion: jobtemplate-2023-09
+name: LONGPROJECTNAME_EPISODE23_SEQUENCE54_SHOT23_REV12_BACKGROUND_CAMERA001
```
edwards-aws, line 26: Would love a better example for this.

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
+````diff
+diff --git a/wiki/2023-09-Template-Schemas.md b/wiki/2023-09-Template-Schemas.md
+index 44d1019..4342177 100644
+--- a/wiki/2023-09-Template-Schemas.md
++++ b/wiki/2023-09-Template-Schemas.md
+@@ -1075,15 +1075,16 @@ command: <CommandString> # @fmtstring[host]
+ args: [ <ArgString>, ... ] # @optional @fmtstring[host]
+ timeout: <posinteger> # @optional
+ cancelation: <CancelationMethod> # @optional
++python | bash | cmd | powershell | node: <DataString> # @optional @fmtstring[host] @extension FEATURE_BUNDLE_1
```
edwards-aws: Would love feedback on this notation, I'm not sold

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
+````diff
+diff --git a/wiki/2023-09-Template-Schemas.md b/wiki/2023-09-Template-Schemas.md
+index 44d1019..74c99f3 100644
+--- a/wiki/2023-09-Template-Schemas.md
++++ b/wiki/2023-09-Template-Schemas.md
+@@ -524,17 +524,19 @@ An `<AmountRequirement>` is the object:
+ ```yaml
+ name: <AmountCapabilityName>
+ min: <nonnegativefloat> # @optional
++min: <nonnegativefloat> | <nonnegativefloatstring>  # @optional @fmtstring @extension FEATURE_BUNDLE_1
```
edwards-aws: This happens a few times in this RFC. We need a good way of saying this type union is only valid with the extension. Not sold on this notation and would love suggestions.

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
+Each addresses issues identified through real-world usage,
+and while individually minor, together they improve the
+developer experience and operational flexibility of OpenJD templates.
+
+### Use Cases Supported
+
+1. **Automated Naming**: When a template is generated programmatically, it's
+natural to concatenate identifiers from the input to get a unique name. The
+current length limits prevent that from working naturally. Identifier names can
+similarly be generated by concatenating identifiers, so we likely want to treat
```
mwiebe: Where you say "we likely want", you need to make a choice so the proposal is concrete. Either the RFC is making that change too, or it's not.

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
+2. **Cross-Platform Compatibility**: Explicit EOL control ensures consistent
+behavior across different operating systems, particularly important for
+embedded scripts and configuration files.
+
+3. **Simplified Template Authoring**: Syntax sugar for common interpreters
+reduces boilerplate and makes templates more readable and maintainable.
+
+4. **Quicker Iteration**: Further parameterization allows for quicker iteration
+times. During experimentation it is quicker to modify a parameter and resubmit a
+job than to edit the source file and resubmit.
```
mwiebe: These values can also depend on the scene or other job inputs, they're not necessarily fixed for a particular job template. Because of that, it's not just for experimentation, it also matters in production.

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
+    precedence.
+
+### 4.1. `<EnvironmentName>`
+
+ A string value subject to the following constraints:
+
+ 1. Allowable characters: Any unicode character except those in the Cc unicode character category.
+ 2. Minimum length: 1 characters.
+-3. Maximum length: 64 characters.
++3. Maximum length: 512 characters.
```
mwiebe: This should say that it's 64 characters unless the FEATURE_BUNDLE_1 extension is enabled and is then 512. Similarly for the other length changes. It would change to just be 512 in the next dated release of the spec, where the extensions become part of the main profile.

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
+
+### Multiple Actions in onRun
+
+The original discussion included a proposal to generalize the `onRun` action
+property in `<StepActions>` to accept either a single `<Action>` as currently
+defined, or a list of `<Action>` objects. The goal was to allow a list of
+sequential actions in the `onRun` of a task, with each command having separate
+timeout and cancelation behavior.
+
+This feature was rejected for inclusion in this RFC because it introduces
```
mwiebe: Since the idea is rejected from the RFC, not in general, I'd suggest to not mention it at all, and rather create a new discussion topic dedicated to it and comment on the existing one that this point is moved there.

mwiebe (changes requested):

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
+-
++
+    The default timeout, if not provided, depends on the action:
+
+    | Action container | Action | Default |
+@@ -1099,6 +1100,137 @@ cancelation: <CancelationMethod> # @optional
+ 4. *cancelation* — If defined, provides details regarding how this action should be canceled. If not provided, then it is
+    treated as though provided with `<CancelationMethodTerminate>`.
+
++5. *python | bash | cmd | powershell | node* - Syntactic sugar for scripts,
```
baxeaz: I really like these

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
+them similarly.
+
+2. **Cross-Platform Compatibility**: Explicit EOL control ensures consistent
+behavior across different operating systems, particularly important for
+embedded scripts and configuration files.
+
+3. **Simplified Template Authoring**: Syntax sugar for common interpreters
+reduces boilerplate and makes templates more readable and maintainable.
+
+4. **Quicker Iteration**: Further parameterization allows for quicker iteration
```
baxeaz: At first read I didn't quite catch which of the changes fall under this - the parametrized action timeout and notifyPeriodInSeconds are described up top under "Enhanced format string support for" but being able to easily resubmit with parameter overrides for these is I think the real motivation?

baxeaz: Great, thanks!  Maybe worth rephrasing in the summary to make that clearer?

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
+ 8. *steps* — A list of the Steps in the Job. See: [&lt;StepTemplate&gt;](#3-steptemplate).
+
+ #### 1.1.1. `<JobName>`
+
+ A [Format String](#73-format-strings) subject to the following constraints:
+
+ 1. Allowed characters: Any unicode character except those in the Cc unicode character category.
+ 2. Minimum length: 1 character.
+-3. Maximum length: 128 characters, after the format string has been resolved.
++3. Maximum length: 512 characters, after the format string has been resolved.
```
epmog: I don't think any of the other names follow the "after the format string has been resolved" wording even if it's true. Maybe we need to fix the spec to make it consistent wording (no behaviour change)?

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
+ specificationVersion: environment-2023-09
+ parameterDefinitions: [ <JobParameterDefinition>, ... ] # @optional
+@@ -450,21 +450,21 @@ Where:
+    take on, and how those parameters' values are combined to produce the Tasks of the Step. Absent this property the Step
+    is run a single time.
+ 7. *script* — The action that is taken by this Step's Tasks when they are run on a Worker host.
+
+ ### 3.1. `<StepName>`
+
+ A string subject to the following constraints:
```
epmog: huh, this doesn't say it's a format string, but it is? Job name does. But some of the others don't. Maybe some consistency miss in the base document?

rfcs/0004-enhanced-limits-and-capabilities.md
epmog: I think this probably works for our use-case here. maybe if we modify the same fields again in another extension (before releasing a new spec revision) then we'll have to do something else.

rfcs/0004-enhanced-limits-and-capabilities.md
edwards-aws: Will do!

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
+
+ 1. Allowed characters:
+    * Unicode alphanumeric characters in the latin character set.
+    * The underscore character ('_').
+ 2. Must start with a letter or underscore character.
+ 3. Minimum length: 1 character
+-4. Maximum length: 64 character
++4. Maximum length: 512 character
```
epmog: nit: characters. Probably just an issue in the base spec

rfcs/0004-enhanced-limits-and-capabilities.md
epmog: 1. Based on the below this option is mutually exclusive with command but nothing else. Can we indicate that in some way here?
2. What would the recommended way a user specifies a version of these tools? Environment set-up?
3. maybe we should sort the options here

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
+
+ 1. *command* — A [Format String](#73-format-strings) containing the name of a runnable command that is run on a Worker
+-   host.
++   host. Cannot be used when *python | bash | cmd | powershell | node* is provided.
+ 2. *args* — An array of [Format Strings](#73-format-strings) that will be passed as arguments to the **command** when the
+    command is run on the host.
+ 3. *timeout* — The positive number of seconds that the command is given to successfully run to completion. A command that
+    does not return before the timeout is canceled and is treated as a failed run.
+-
++
```
epmog: accidental inclusion? or what's changed

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
++          - name: <implicitly generated file>
++            filename: "<implicitly generated file>.py"
++            type: TEXT
++            data:
++                # python code here
++      ```
++
++   2. *bash* - Implicitly creates a Bash embedded file,
++   `command` becomes `bash`, and `args` get prefixed with the implicitly
++   generated file. The file extension is `.sh`.
```
epmog: Should it also be runnable? I guess only if it can be referenced outside this command.

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
+    treated as though provided with `<CancelationMethodTerminate>`.
+
++5. *python | bash | cmd | powershell | node* - Syntactic sugar for scripts,
++removes some commonly needed boilerplate for the given script interpreter.
++Removes the need for the `command` property. Implicitly prefixes the `args` property.
++Removes the need for the `embeddedFiles` property.
++Cannot be used when `command` is provided. Requires the `FEATURE_BUNDLE_1` extension.
++   1. *python* - Implicitly creates a Python embedded file,
++   `command` becomes `python`, and `args` get prefixed with the implicitly
++   generated file. The file extension is `.py`.
```
epmog: Curious, do we need a way to refer to that embedded file elsewhere?

epmog: Seems that this sort of depends on if we want to be able to refer to the same embedded file in an environment OnEnter/OnExit, since currently there is only OnRun for task runs. Maybe there's an an implicit embedded file we can expose via the action or top-level name?

Otherwise that's a use-case where maybe we acknowledge that the syntactic sugar doesn't make sense.

rfcs/0004-enhanced-limits-and-capabilities.md
mwiebe: Step names are structural in addition to being descriptive. For example step dependencies are defined by this name. Making them be a format string doesn't feel like it fits with that purpose to me. By comparison, the job name's primary purpose is to be descriptive.

epmog: Yea, my comment doesn't make sense if it's not a format string, ignore me I thought they were :)

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,655 @@
...
+ 4. *cancelation* — If defined, provides details regarding how this action should be canceled. If not provided, then it is
+    treated as though provided with `<CancelationMethodTerminate>`.
+
++5. *python | bash | cmd | powershell | node* - Syntactic sugar for scripts,
++removes some commonly needed boilerplate for the given script interpreter.
++Removes the need for the `command` property. Implicitly prefixes the `args` property.
++Removes the need for the `embeddedFiles` property.
++Cannot be used when `command` is provided. Requires the `FEATURE_BUNDLE_1` extension.
++   1. *python* - Implicitly creates a Python embedded file,
++   `command` becomes `python`, and `args` get prefixed with the implicitly
```
jusiskin: I'm wondering about an implementation detail specific to the `python` one. Some Linux distributions Python installs do not provide a `python` command on the `PATH`. They now only provide `python3`.

For example, on Ubuntu we see:

```sh
$ uname -a
Linux ip-172-31-11-201 6.14.0-1011-aws #11~24.04.1-Ubuntu SMP Fri Aug  1 02:07:25 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
$ python
sh: 2: python: not found
$ python3
Python 3.12.3 (main, Jun 18 2025, 17:59:45) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>
```

We have a couple of options here:

1. Specify that OpenJD-compatible session implementations account for this (determine whether `python`/`python3` is available, and use it)
2. Provide a `python` and `python3` option here to let OpenJD template authors specify the command
3. Leave this as-is, but maybe expand in the spec that worker hosts have the `python` command available. We could maybe document that debian-based distros can use the [`python3-is-python` package](https://packages.debian.org/search?suite=all&section=all&arch=any&searchon=names&keywords=python-is-python3) to achieve this.

rfcs/0004-enhanced-limits-and-capabilities.md
edwards-aws: 1. Open to suggestions for notation here. We could probably add new notation for mutual exclusion.
2. This is the wording we have for command currently 
   > A [Format String](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md#73-format-strings) containing the name of a runnable command that is run on a Worker host.

   I can be more explicit here and say "X is expected to be in the system PATH"
3. 👍

rfcs/0004-enhanced-limits-and-capabilities.md
mwiebe: There's also the converse issue, when there is `python` but not `python3`. I think it's important to nudge towards portable job templates, and solutions that avoid using a `python3` command are better for portability.

edwards-aws: I think we can keep it as is and add more clarification in the description that states something like

```
python (all lowercase) is expected to be a runnable command on a Worker host.
```

and similar for all other interpreters.

epmog: maybe that messaging adjusted with "available to the user running the command". It's definitely something folks can get tripped up on for PATH, aliases, etc. if they only modify their user and now who runs the command

rfcs/0004-enhanced-limits-and-capabilities.md
edwards-aws: I don't think we want it to be runnable. As it's implicitly run with `bash`, so there's no need for the executable permission to be set. It seems to me that making the file runnable is something that should be explicit? I could be off base there though.

rfcs/0004-enhanced-limits-and-capabilities.md
mwiebe: Is it more "available in the runtime environment" compared to "available to the user"? The command could be provided by a step environment, a job environment, an externally defined environment, the user PATH configuration, or the system PATH configuration.

rfcs/0004-enhanced-limits-and-capabilities.md
edwards-aws: Moved to this [discussion](https://github.com/OpenJobDescription/openjd-specifications/discussions/97). Will remove this section in the next revision.

wiki/2023-09-Template-Schemas.md
```diff
@@ -1223,6 +1383,9 @@ data: <DataString> # @fmtstring[host]
 * *runnable* — A boolean `True` value indicates that the file written to disk should have its execute-permissions set
   to true. Defaults to `False` if not provided.
 * *data* — The string data that will be written to the file exactly as it appears.
+* *endOfLine* — The line endings that the embedded file will have when written
+to disk. If `AUTO` the embedded file will have the default line endings of the
+host operating system. Default is `AUTO`. Requires the `FEATURE_BUNDLE_1` extension.
```
mwiebe, line 1259: Looks like these two lines need to be indented to continue the bullet.

wiki/2023-09-Template-Schemas.md
mwiebe: You'll want to rebase to include https://github.com/OpenJobDescription/openjd-specifications/commit/b9a0eda1b7b9bd4c471a8b99bda6a8b7f6203c7c in your history, then add the extension name to this list.

wiki/2023-09-Template-Schemas.md
mwiebe: Hmm, I guess this formats well in the rendered view, it's just a question of how it looks as a text file. Not important to change.

rfcs/0004-enhanced-limits-and-capabilities.md
```diff
@@ -0,0 +1,223 @@
...
+          echo 'Running'
+        endOfLine: LF
+```
+
+### Script Interpreter Syntax Sugar
+
+```yaml
+steps:
+- name: RunPythonScript
+  script:
```
mwiebe: Would look nice to skip the 'actions' and 'onRun' nested objects when the only thing inside is a 'python' or other syntax sugar script. Do you think that's unambiguous/clear enough and doable in this rfc?

edwards-aws: I think that would be nice, let me play around with the spec a bit to see how that would look in practice.

edwards-aws: So I can think of a couple ideas.

1)
```yaml
actions: <StepActions>
embeddedFiles: [ <EmbeddedFile>, ... ] # @optional
python | bash | cmd | powershell | node: <DataString> # @optional @fmtstring[host] @incompatible command embeddedFiles @extension FEATURE_BUNDLE_1
```
in this you can optionally set the syntax sugar script at the `<StepScript>` level. If that is done you can't specify `actions` or `embeddedFiles`. Additionally you can't specify things like `args`, or timeouts. All defaults will be used.

This would be nice for quick, super simple scripts.

2) We create a new object called `<SimpleAction>` (let me know if you have a better name). This looks the exact same as an `Action`, but with all of the syntactic sugar included and no command.

```yaml
actions: <StepActions>
embeddedFiles: [ <EmbeddedFile>, ... ] # @optional
python | bash | cmd | powershell | node: <SimpleAction> # @optional @fmtstring[host] @incompatible command embeddedFiles @extension FEATURE_BUNDLE_1
```

```yaml
script: <DataString> # @fmtstring[host]
args: [ <ArgString>, ... ] # @optional @fmtstring[host]
timeout: <posinteger> # @optional
timeout: <posinteger> | <posintstring> # @optional @fmtstring
cancelation: <CancelationMethod> # @optional
```

This new object also gives us the option of using it for things like environment scripts.

Let me know what you think

mwiebe: I like the latter idea, that seems cleaner. If we want to apply it to the environment script case, the simple action would become onEnter and there would be no onExit, right? Similarly in the future if/when we add more actions than onRun to the step.

edwards-aws: I think that makes sense.

rfcs/0004-enhanced-limits-and-capabilities.md
edwards-aws: See the conversation here https://github.com/OpenJobDescription/openjd-specifications/pull/95#discussion_r2621725217

I think that mostly addresses this.

edwards-aws:
Please note https://github.com/OpenJobDescription/openjd-specifications/pull/95 is now in final-comments - if there are any other concerns please let me know!

mwiebe (approved):

Issue #112, opened 2026-02-18
RFC: Add EXPR extension for a glue expression language in format strings


mwiebe:
* **Pull Request**: https://github.com/OpenJobDescription/openjd-specifications/pull/113
* **Discussion Thread(s)**:
    * https://github.com/OpenJobDescription/openjd-specifications/discussions/79

## Description

This is a proposal split into 3 RFCs to make each one better-defined and easier to review.

### RFC 5 - Expression Language

Open Job Description templates need a flexible way to customize job structure and express glue transformations between different interfaces. Schedulers must understand job structure without running tasks to determine it, so need the ability to evaluate it in an isolated, secure, and bounded context. We propose a domain-specific expression language to provide this flexibility. The language we define has a type system and set of operations rich enough to cover the diverse use cases identified in community discussions, with well-defined, bounded expression evaluation semantics.

### RFC 6 - Expression Function Library

This RFC defines the operators and built-in functions for the expression language introduced in RFC 0005. By separating the function library from the core language specification, we enable independent evaluation and evolution of these concerns. The library provides arithmetic, string, list, path, and serialization operations sufficient for common job template use cases.

### RFC 7 - Extended Parameter Types

This RFC extends the job parameter type system with boolean and list types, and makes type names case-insensitive. These additions enable template authors to express common patterns more naturally and align type syntax with surrounding YAML/JSON conventions.

### Working Prototype Implementation

I've created a fully-functioning prototype implementation across openjd-model, openjd-sessions, and openjd-cli.

The RFC and all the prototype implementation were created with the assistance of AI ([Kiro CLI](https://kiro.dev/cli/) using Claude Opus 4.5 and 4.6).

## Roles

| Role | User
| ---- | ----
| Proposed By | @mwiebe
| Author(s)   | @mwiebe

## Workflow

- [x] Tracking issue created (label: `rfc/proposed`)
- [x] RFC pull request submitted and ready for discussion (label: `rfc/exploring`)
- [x] Last call for comments (labels: `rfc/exploring` and `rfc/final-comments`)
- [x] Accepted and merged RFC pull request (label: `rfc/accepted-future`)
- [x] Green-light for inclusion in a draft specification, and the author is creating and iterating on pull requests (label: `rfc/accepted-draft`)
- [x] Pull requests are merged in to a draft specification (label: `rfc/accepted-staged`)
- [ ] Officially published in a non-draft revision of the specification (label: `rfc/released`)

[...]

PR #113, opened 2026-02-18
feat(RFC): An expression language for format strings

mwiebe:
## PR for RFC

**Tracking Issue**: https://github.com/OpenJobDescription/openjd-specifications/issues/112

This is a request for comments about the proposed RFC 5, 6, and 7.

This is a set of 3 RFCs to propose an expression language to use in template format strings based on a subset of the Python language. The three proposals are:

RFC 5 - Expression Language
RFC 6 - Expression Function Library
RFC 7 - Extended Parameter Types

I have a fully-functioning prototype implementation across these branches:

* https://github.com/mwiebe/openjd-model-for-python/tree/expr
* https://github.com/mwiebe/openjd-sessions-for-python/tree/expr
* https://github.com/mwiebe/openjd-cli/tree/expr

This branch contains edits to show how things look when using the extension:

* https://github.com/mwiebe/deadline-cloud-samples/tree/expr

This branch in my fork of OpenCue is an experiment implementing a pyoutline backend for OpenJD:

* https://github.com/mwiebe/OpenCue/tree/openjd-prototype

---


crowecawcaw:
I'm curious how you'd compare the approach of adding new syntax to using external tools for features that OpenJD doesn't cover yet. A couple ways external tools could be integrated:
- Using a high level language like Python to generate the job templates, like what Deadline Cloud submitters often do
- Adding some sort of hook or preprocessor option to job templates that allows an arbitrary process / custom tool to handle expressions or logic

I'd think at a high level, the new syntax allows more expressive OpenJD templates to still be portable since we wouldn't need an external tool (like Python) to preprocess the job template. Also it'd let us build more features off of OpenJD like conditionally hiding job submitter UI elements. But the cost is adding complexity to OpenJD that alternative implementations need to support both for parsing job templates and for interacting with them (e.g. building an OpenJD-based submitter).

With agents getting better, the complexity cost is probably lower than it was a year ago. I'm also curious about your experience adding support to OpenCue.

The new expression syntax seems like the right approach to me. I'm asking because I'm sure you've thought about the tradeoffs and I'd like to understand them more.

mwiebe:
> I'm curious how you'd compare the approach of adding new syntax to using external tools for features that OpenJD doesn't cover yet. A couple ways external tools could be integrated:
> 
> * Using a high level language like Python to generate the job templates, like what Deadline Cloud submitters often do

When we first made the spec, we also started the pattern in Deadline Cloud submitters because we weren't able to fit in enough features to represent every variation the submitters need. That pattern uses code to generate a template that has mostly has a general template with parameters for the specific submission, but hardcodes some parts of it. The result is that the template isn't then a general reusable entity - it is just for the specific submission. When we can express everything in the template and have the submitter just set parameters, the result is available to read and re-apply to other scenes directly.

> * Adding some sort of hook or preprocessor option to job templates that allows an arbitrary process / custom tool to handle expressions or logic

This approach adds constraints on the environment you're submitting a job template from, that an extension to the spec does not. Think of job submissions from the desktop vs from a browser, with and without a file system, etc. Expressions inside the template don't have to be understood at the submission context, but a preprocessor forces the submission context to have the ability to run the preprocessor commands.

> I'd think at a high level, the new syntax allows more expressive OpenJD templates to still be portable since we wouldn't need an external tool (like Python) to preprocess the job template. Also it'd let us build more features off of OpenJD like conditionally hiding job submitter UI elements. But the cost is adding complexity to OpenJD that alternative implementations need to support both for parsing job templates and for interacting with them (e.g. building an OpenJD-based submitter).

Yes, I think this expression language hits the [design tenets](https://github.com/OpenJobDescription/openjd-specifications/wiki/Design-Tenets) well.

1. Portable: the language is defined independent of the submission context or implementation details of a scheduler.
2. Expressive: It adopts familiar and powerful Python syntax and starts with a broad function library.
3. Human readable and writable: Python is familiar, and I've been experimenting with it myself to help validate this.
4. Tooling parseable: The openjd.expr library interface exposes the symbol references, called functions, and result type for a parsed expression, letting tools understand variable dependencies within the template and whether the template accesses path mapping.

> With agents getting better, the complexity cost is probably lower than it was a year ago. I'm also curious about your experience adding support to OpenCue.

The OpenCue support I wrote is a very incomplete prototype - it is adding openjd support to the pyoutline library, which is for defining jobs to submit to OpenCue. I think I got the basic structure right, but it needs a lot of validation before it would be ready to use. The opposite, of running an OpenJD job in OpenCue, is not something I've looked at yet.

> The new expression syntax seems like the right approach to me. I'm asking because I'm sure you've thought about the tradeoffs and I'd like to understand them more.

Thanks!

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2321 @@
...
+| `FLOAT` | `float` |
+| `PATH` | `path` |
+| `BOOL` | `bool` |
+| `RANGE_EXPR` | `range_expr` |
+| `LIST[STRING]` | `list[string]` |
+| `LIST[INT]` | `list[int]` |
+| `LIST[FLOAT]` | `list[float]` |
+| `LIST[PATH]` | `list[path]` |
+| `LIST[BOOL]` | `list[bool]` |
+| `LIST[LIST[INT]]` | `list[list[int]]` |
```
crowecawcaw: These look the same. Are we able to collapse the concepts?

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2321 @@
...
+- **Host context (SESSION/TASK scope)**: Extended library including `apply_path_mapping()`
+
+This allows the type checker to correctly validate that `apply_path_mapping()` is only used
+in contexts where it will be available at runtime.
+
+### Progressive Expression Evaluation
+
+Expressions are checked and evaluated progressively as more information becomes available.
+At each stage, known values are evaluated concretely while unknown values propagate as
+`unknown[T]`, catching as many errors as possible with the information available:
```
crowecawcaw: Might be useful to make the existing stages more explicit in the base spec so it's clear where this fits in.

I think these stages would also benefit from a diagram or table or similar.

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2321 @@
...
+          args: ["-p", "{{ work_dir }}"]
+```
+
+## Specification
+
+### Extension Name
+
+`EXPR`
+
+### Extended Format String Grammar
```
crowecawcaw, line 557: I think it'd help to have a high level summary up here with a shorter, less formal description of the extension, a quick glossary of concepts, and some diagrams.

Some ideas for things to put in the summary:
- the validation/evaluation phases
- the type system vs parameter types
- symbols, ast, expression parsing, and how and when they fit together

rfcs/0005-expression-language.md
mwiebe: I collapsed this and the task parameter types.

rfcs/0007-extend-parameter-types.md
```diff
@@ -0,0 +1,481 @@
...
+    type: int          # lowercase now valid
+  - name: OutputDir
+    type: Path         # mixed case now valid
+```
+
+## Motivation
+
+The current parameter type system has limitations:
+
+1. **No boolean type** - Users must use `STRING` with `allowedValues: ["true", "false"]` to
```
leongdl, line 71: I love this! Came across this in 3dsmax quite often for parameter passing. 

Given backwards compatibility, would be awesome to accept union[bool|string] type of interface too, but that's probably overkill.

rfcs/0005-expression-language.md
mwiebe: Added a table here:

https://github.com/mwiebe/openjd-specifications/blob/6064bb6bcde54b057f03cf3deedb9c6a817902db/wiki/2023-09-Template-Schemas.md#74-template-processing-stages

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2322 @@
...
+args:
+  - "--input"
+  - "{{Param.InputFile}}"
+  - "{{ '--verbose' if Param.Verbose else null }}"  # Dropped when false
+  - "{{ ['--quality', Param.Quality] if Param.Quality > 0 else null }}"  # Flattened or dropped
+```
+
+### 6. Let Bindings for Common Sub-expressions
+
+To avoid repeating complex expressions across multiple fields, `let` bindings can be added
```
leongdl, line 416: Curious if we would want to consider a job level `let` syntax to compute job wide variables based on input params?

rfcs/0007-extend-parameter-types.md
```diff
@@ -0,0 +1,481 @@
...
+
+- `INT`, `Int`, `int`
+- `FLOAT`, `Float`, `float`
+- `STRING`, `String`, `string`
+- `PATH`, `Path`, `path`
+- `BOOL`, `Bool`, `bool`
+- `RANGE_EXPR`, `Range_Expr`, `range_expr`
+- `LIST[STRING]`, `List[String]`, `list[string]`
+- `LIST[INT]`, `List[Int]`, `list[int]`
+- `LIST[LIST[INT]]`, `List[List[Int]]`, `list[list[int]]`
```
leongdl, line 96: Two levels deep — list[list[T]] is the max. No list[list[list[T]]].

From the RFC's type constraints on list[T]:

T cannot be S?. A None/null value inside a list literal is an error.
T can be list[S], but cannot be nested a third time, so S cannot be list[U].
So these are valid:

list[int] — [1, 2, 3]
list[list[int]] — [[1, 2], [3, 4]]
list[list[string]] — [["a", "b"], ["c"]]
And this is not:

list[list[list[int]]] — error
This keeps the type system simple and memory tracking straightforward. The main use case for list[list[T]] is things like [['-e', env_var] for env_var in Task.Environment] which produces a nested list that gets flattened into args. A third level of nesting doesn't have a practical use case in job templates.

Kiro is pretty smart and able to deduce this!

rfcs/0007-extend-parameter-types.md
leongdl: Was also thinking about `tuples`, eg {a, 1, true}, but looks like the syntax of let groups would handle it well !

rfcs/0007-extend-parameter-types.md
mwiebe: An adjacency list for task dependencies is another use case, for `list[list[int]]`.

rfcs/0005-expression-language.md
mwiebe: Yes, I think it's worth considering that together with the environment variable idea https://github.com/OpenJobDescription/openjd-specifications/discussions/104 and variable data flow https://github.com/OpenJobDescription/openjd-specifications/discussions/111

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2322 @@
...
+### Memory-Bounded Evaluation
+
+Expression evaluation must operate within bounded memory to support constrained execution
+environments and predictable resource usage when evaluating many expressions concurrently
+across threads.
+
+Implementations accept an optional `memory_limit` parameter (default: 10MB recommended).
+During evaluation, the evaluator tracks the memory size of live values—incrementing when
+values are created, decrementing when intermediate values are consumed by operations.
+If current memory exceeds the limit at any point, evaluation fails with an error.
```
crowecawcaw: What do you think about softening this requirement for the spec? Something like, "Implementations SHOULD add memory and runtime guards to prevent bad job templates from crashing the interpreter." Another implementation might want to track memory differently or may do something like cap the number parameter values, if it's easier.

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2322 @@
...
+
+Examples:
+- `[1, 2, 3, 4, 5][1:4]` returns `[2, 3, 4]`
+- `"hello"[1:4]` returns `"ell"`
+- `path("/a/b/c/d").parts[1:]` returns `["a", "b", "c", "d"]`
+- `range_expr("1-10")[::2]` returns `[1, 3, 5, 7, 9]`
+
+### Conditional Expression Semantics
+
+The conditional expression `<true_value> if <condition> else <false_value>`:
```
epmog, line 1797: In my mind, I see this as something that's going to be written frequently when dealing with `null`. Maybe now's the right time to think about what a more compact version looks like?

Can leverage something like Python's `or` syntax:

```python
value or "fallback"
```

Or do something like jinja, etc.

```jinja
{{ value | default("fallback") }}
```

Let me know your thoughts

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2371 @@
...
+   type checking and implicit coercion (e.g., `int` → `float`). The evaluation rules are
+   not a subset of Python semantics, but are intended to be simple and familiar.
+
+4. **Result** — A typed value that is converted to a string for insertion into the format
+   string, or used directly in `let` bindings and parameter space definitions.
+
+### Evaluation Stages
+
+Not all values are known at the same time. The evaluator handles this through three stages,
+using placeholders (`unknown[T]`) to perform type checking on values not yet available:
```
crowecawcaw: In TypeScript's type system (just as a point of comparison), `unknown` means the type is unknown  (e.g. when loading data from a JSON blob). For me, that feels more intuitive than an `unknown` type to mean an unknown value.

What do you think about a different term like `unresolved` or `token` to indicate placeholders? For the goal of making the term less ambiguous about what's unknown.

Might be more of a preference, so just an idea.

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2371 @@
...
+| Expression Type | Parameter Type(s) | Description |
+|----------------|-------------------|-------------|
+| `int` | `INT` | Integer values |
+| `float` | `FLOAT` | Floating-point values |
+| `string` | `STRING` | Text values |
+| `path` | `PATH` | Filesystem paths or URIs |
+| `bool` | `BOOL` | True/false values |
+| `range_expr` | `RANGE_EXPR` | Compact integer range (e.g. `"1-10,20"`) |
+| `list[T]` | `LIST[T]` | Ordered collection of values |
+| `nulltype` (null) | — | Absence of a value (for conditional omission) |
```
crowecawcaw: Could we abbreviate this to something like "expressions have a type and evaluate to the equivalent of a parameter".

rfcs/0006-expression-function-library.md
```diff
@@ -0,0 +1,907 @@
...
+#### Logical Operators
+
+| Signature | Description |
+|-----------|-------------|
+| `a and b` | If `a` is `null` or `false`, return `a`; otherwise evaluate and return `b` (short-circuit) |
+| `a or b` | If `a` is `null` or `false`, evaluate and return `b`; otherwise return `a` (short-circuit) |
+| `__not__(a: bool) -> bool` | `not a` logical NOT |
+
+The `and` and `or` operators are value-returning: they return one of their operands, not
+necessarily a `bool`. Only `null` and `false` are considered falsy — unlike Python, values
```
crowecawcaw: This break from Python feels like it could be confusing since the rest of the syntax is basically Python (though the motivation for a nice fallback mechanism makes sense).

What do you think about either:
- using Python's semantics. People and LLMs understand them already
- or making or/and return bool always (so forcing a more explicit `X if X else Y` pattern)

rfcs/0006-expression-function-library.md
```diff
@@ -0,0 +1,907 @@
...
+like `0`, `""`, and `[]` are not falsy. This makes `or` useful as a null-coalescing operator:
+
+```python
+Param.X or "fallback"              # returns Param.X if not null/false, else "fallback"
+Param.X and Param.X.stem           # returns null if X is null/false, else X.stem
```
crowecawcaw: This returns `false` if X is `false` right?

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2371 @@
...
+    script:
+      let:
+        - work_dir = Param.OutputDir / 'work'
+      actions:
+        onEnter:
+          command: mkdir
+          args: ["-p", "{{ work_dir }}"]
+```
+
+## Specification
```
crowecawcaw, line 551: Do we need to specify the order of operations? Or maybe just spec that it follows Python? Might be noted somewhere and I missed it.

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2371 @@
...
+| Symbol | Source | Example |
+|--------|--------|---------|
+| `Param.<name>` | Job parameter values (path-mapped for PATH) | `Param.OutputDir` |
+| `RawParam.<name>` | Job parameter values (unmapped) | `RawParam.InputFile` |
+| `Task.Param.<name>` | Task parameter values | `Task.Param.Frame` |
+| `Job.Name` | The resolved job name | `Job.Name` |
+| `Step.Name` | The current step's name | `Step.Name` |
+| `Session.*` | Runtime session info | `Session.WorkingDirectory` |
+| `Task.File.<name>` / `Env.File.<name>` | Embedded file locations | `Task.File.Script` |
+| User-defined (`let`) | Let bindings | `output_path` |
```
crowecawcaw: This is the same as before right? let expressions are new but otherwise it looks like same.

crowecawcaw: Looks like `Job.Name` is new but some others are existing. Is there a list of new symbols being added?

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2371 @@
...
+
+- **Implicit line continuation** - Multi-line expressions are supported without requiring
+  backslash (`\`) continuation or enclosing parentheses. See
+  [Appendix D: Implicit Line Continuation](#appendix-d-implicit-line-continuation) for
+  implementation details.
+
+See [Appendix B: Language Syntax Choice](#appendix-b-language-syntax-choice) and
+[Appendix A: Backward Compatibility Analysis](#appendix-a-backward-compatibility-analysis) for rationale.
+
+### 2. Float Value Pass-Through
```
crowecawcaw, line 321: This feature seems nice but has adds cognitive load for being different from Python. Is it needed? Or what motivates it?

rfcs/0007-extend-parameter-types.md
mwiebe: Think about it in terms of future extension. If there's something you can't express now, will extending to support that be a natural additive extension of what we've designed? If not, can we tweak the design to make that future extension easier?

rfcs/0005-expression-language.md
mwiebe: I was thinking along the opposite lines - to extend this with an additional operation count, and put a cap on that too. Can you give some more concrete examples of what you're thinking? How different would the memory tracking be that it can't fit in this proposed interface?

rfcs/0005-expression-language.md
mwiebe: Have a look at https://github.com/mwiebe/openjd-specifications/blob/expr/rfcs/0005-expression-language.md#ast-transformation-to-uniform-function-calls where I've defined `or` to be the language's null-coalescing operator. What do you think? This is different from Python, but when we add the ability for job parameters to have an optional type, it would let `Param.OptionalValue or (calculated default)` work.

rfcs/0006-expression-function-library.md
mwiebe: I did this based on @epmog's feedback above (see https://github.com/OpenJobDescription/openjd-specifications/pull/113#discussion_r2885262644). This choice lets you express null-coalescing very easily, which in Python syntax is not possible without a longer expression.

Let's extend the example from that comment. So `Area` is a job parameter with `int?` type (assuming an extension to optional job parameters), while `Width` and `Height` are integers. The expression `Param.Area or (Param.Width * Param.Height)` says "use the area if provided, otherwise calculate from width and height." To use Python semantics, you have to say `Param.Width * Param.Height if Param.Area == None else Param.Area`, otherwise you run into trouble with the Param.Area == 0 case.

rfcs/0006-expression-function-library.md
mwiebe: Correct, that comment is wrong

rfcs/0005-expression-language.md
mwiebe: This is an extension of a property we put in the first release of OpenJD - if you provide a float job parameter value like "3.50", it will get substituted verbatim when substituted into the job via a format string. The motivation was to give control to the template author and submitter over how the strings get put together in the templating, to give a better "pass through just works" experience. This rule extends that into the expression language in the most natural way I could think of.

rfcs/0005-expression-language.md
mwiebe: This type system is a superset of job/task parameters, and implementations of it will need to follow it precisely to get all the details right. I think the table is worth keeping here. EDIT - sorry, it does make sense to summarize it in the overview and leave the dense material for the later specification.

rfcs/0005-expression-language.md
crowecawcaw: (caveat: heavy on vibes, light on hard data) 

I agree implementations will want/need guarantees around memory and runtime. I would have a hard time entirely trusting a value counter for enforcing it well though. What if the counter misses something? Or what if my implementation uses more memory than the spec assigns to a value? Putting this in the spec requires this specific implementation.

An alternative implementation could be running the job in a subprocess with a hard timeout and hard memory cap. I'd guess it'd be easier to implement and I'd trust it more. The downside would be that the template author can't predict whether a template will hit a limit or not before trying it, but I think in practice, limits will be very high so only malicious or malformed templates would be rejected. Any currently implementations can still choose their own memory limit, so there's still some dependency on the implementation.

Value/operation counting seems fine for our current implementation but it feels like it could be moved out of the spec.

rfcs/0005-expression-language.md
mwiebe: I see, so Python's `Any` and TypeScript's `unknown` are the same. I went with `any` to represent "this value can have any type" and `unknown` to represent "this is a concrete value, but we don't know its value and we only have partial information about the type."

rfcs/0006-expression-function-library.md
mwiebe: Have a read of https://numpy.org/neps/nep-0026-missing-data-summary.html if you want to dive deeper into a similar-ish topic! There the missing values are for calculation, here they are for type checking.

rfcs/0005-expression-language.md
mwiebe: Since this is the overview, I think maybe 2 examples should be enough here.

rfcs/0005-expression-language.md
mwiebe: Maybe at the overview level it shouldn't mention `unknown[T]`? That's for the details

rfcs/0005-expression-language.md
mwiebe: Which order of operations do you mean? Your comment is on the top level specification header.

rfcs/0005-expression-language.md
mwiebe: > What if the counter misses something?

This would be a bug to fix. All the functions and operations are defined in the spec, so it's a closed system.

> Or what if my implementation uses more memory than the spec assigns to a value?

Something to fix in the spec, I think there are some left over hardcoded numbers, the memory usage should be implementation-defined to represent its actual memory usage.

It could be moved out of the spec, but I think it's valuable to have it in the API for openjd.expr. Can you expand on why you would want to move it out of the spec?

rfcs/0005-expression-language.md
mwiebe: Shortened for my next push.

rfcs/0005-expression-language.md
mwiebe: Shortened.

rfcs/0005-expression-language.md
mwiebe: I've tightened up the language here. Want to start a naming discussion about `unknown` further below in the full spec for the types?

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2368 @@
...
+| `string` | String values (name matches the job parameter type, not Python `str`) |
+| `path` | Filesystem path values |
+| `range_expr` | Range expression string conforming to `<IntRangeExpr>` grammar |
+| `T?` | The type is `T`, but can also be `None`/`null` |
+| `nulltype` | The type is like Python `NoneType`, its value can only be `None`/`null` |
+| `list[T]` | Ordered list of values of type `T` (see below for constraints on `T`) |
+| `list[nulltype]` | Empty list of values `[]`, where we don't know what `T` should be. |
+| `T1 \| T2` | The type may be either `T1` or `T2` |
+| `noreturn` | Bottom type for functions that never return a value (e.g., `fail()`). In unions, collapses to nothing: `T \| noreturn` simplifies to `T`. |
+| `unknown[T]` | Value unknown, but satisfies constraint `T` (type-checking only) |
```
crowecawcaw: Idea: maybe choose a different word than `unknown` that implies the value is not yet know and which does not imply the type is ambiguous. Where I'm coming from: in TypeScript which I've used a lot, `unknown` means that a type is unknown.

Ideas:
 - `unresolved`
 - `pending` overlaps with Deadline Cloud's statuses, but those aren't part of OJD that I know of
 - `token` as in, this is not the real value, it's a placeholder
 - `future`
 - `missing`, but I don't like it. sounds broken to me
- `not_avail`, offline suggestion from @mwiebe 

I like `pending` best but it's a weak preference. `unknown` isn't bad either if we still prefer it.

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2368 @@
...
+
+### How Expressions Are Processed
+
+Expressions appear inside `{{ }}` in format strings and in `let` bindings. Processing follows
+these steps:
+
+```
+Template format string: "render --frames {{Param.Start}}-{{Param.End}}"
+                                         ├─────────────┘ ├───────────┘
+                                         expression 1     expression 2
```
crowecawcaw: Nit: it'd be a better illustration if at least one of these expressions had some new features like arithmetic in them

rfcs/0005-expression-language.md
mwiebe: `unresolved` is my top pick from that list. In the implementation there is a `resolve()` function being used, so that lines up.

rfcs/0005-expression-language.md
mwiebe: I fixed up the language so it's implementation-defined, and changed the default to 100MB. We should test with some actual expressions at the limits of what we expect to verify that's good.

rfcs/0005-expression-language.md
mwiebe: Changed to `unresolved`.

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2433 @@
...
+
+**Value Size Calculation:**
+
+The size of a value is implementation-defined. Implementations should try to match the
+actual memory usage of each value as closely as practical in their language and runtime.
+
+**Memory Tracking During Evaluation:**
+
+When evaluating a binary operation like `left + right`:
+1. Evaluate `left` → add `size(left)` to current memory
```
crowecawcaw, line 1146: If memory constraints are up to the implementation, should we punt the memory tracking details from the RFC to the model PR?

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2433 @@
...
+checked as each element is added, failing once the accumulated list size exceeds it.
+
+### Operation-Bounded Evaluation
+
+Expression evaluation must operate within a bounded number of operations to prevent
+unbounded computation from deeply nested or combinatorially explosive expressions.
+
+Implementations accept an optional `operation_limit` parameter (default: 10 million
+recommended). During evaluation, the evaluator maintains a running operation count.
+If the count exceeds the limit at any point, evaluation fails with an error.
```
crowecawcaw, line 1171: I have similar concerns here as with the memory counting idea. Counting operations adds complexity to the spec and constrains implementations, but it still doesn't give strong guarantees against runtime timeouts. For example, 10M regex operations might take an unreasonable amount of time.

I think the spec should be open to an implementation capping memory and runtime with a cgroup and a timeout which would offer firmer guarantees. Or run the job parser in a Lambda function with a timeout.

rfcs/0005-expression-language.md
```diff
@@ -0,0 +1,2433 @@
...
+Note: Float values use 64-bit IEEE 754 representation and are subject to standard floating-point
+precision (e.g., `0.1 + 0.2` produces `0.30000000000000004`, not `0.3`). The expression language
+does not produce negative zero, infinity, or NaN:
+
+- Negative zero (`-0.0`) is normalized to `0.0` after every operation.
+- Operations that would produce infinity (e.g., `1e300 * 1e300`) are errors.
+- Operations that would produce NaN (e.g., `0.0 / 0.0`) are errors.
+- `float('inf')`, `float('nan')`, and `float('-inf')` are errors.
+
+### 3. 64-bit Signed Integer Type
```
crowecawcaw, line 342: I think it'd help to add a bullet list summary in the RFC of divergences from Python's expressions. It'd  be useful for checking design choices and as a list of things to double check in an implementation.

rfcs/0005-expression-language.md
mwiebe: The size of the values are implementation-defined. A conforming implementation still needs to track it. I'm not sure why we would want to exclude this from the spec - maybe I can describe this as illustrative, and the key is that any implementation needs to track in a reasonable way.

rfcs/0005-expression-language.md
mwiebe: I see what I've proposed here as stronger and more portable than something based on cgroups and timeouts. There's a lot of operating system variation for that kind of mechanism, but what's specified here is quite simple to implement everywhere.

mwiebe: Re: the regexes, that's a fair point for the current Python implementation. Rust's regex doesn't use unbounded backtracking per https://docs.rs/regex/latest/regex/#untrusted-input, and I'm interested in getting a Rust implementation of this spec. I see no obstacles for either the memory or operation tracking in a Rust port.

rfcs/0005-expression-language.md
mwiebe: See https://github.com/OpenJobDescription/openjd-specifications/compare/c1801ab9bb5c01158fb4d3a5b6d9fc349a02742a..1f02cf986fc275d0af640e684ed498593f7c5a5d

rfcs/0005-expression-language.md
crowecawcaw: I had in mind the order for operations like `or` and `==`. Looks like it's defined formally in the grammar though. And since it matches Python, maybe there's no need to say anything more. If we did want to mention it, I think just a line like "Operator precedence follows Python's operator precedence" would cover it.

crowecawcaw (dismissed):

mwiebe:
Following the RFC process, I've updated the status of https://github.com/OpenJobDescription/openjd-specifications/issues/112 to rfc/final-comments. Please take a look if you have any comments outstanding or want to help with polishing.

leongdl (approved):

Issue #124, opened 2026-04-08
RFC: JSON Embedded Files


crowecawcaw:
* **Pull Request**: (Add the URL to the pull request here.)
* **Discussion Thread(s)**:
  * https://github.com/OpenJobDescription/openjd-specifications/discussions/122

## Description

Add a new type: "JSON" embedded file variant where data is a structured object instead of raw text. The runtime resolves format string expressions at the value level within the parsed data  structure, then serializes the result to JSON — eliminating the class of quoting and escaping bugs that occur when parameter values containing special characters are substituted into TEXT embedded files.        

## Roles

| Role | User
| ---- | ----
| Proposed By | @crowecawcaw 
| Author(s)   | @crowecawcaw 

## Workflow

- [x] Tracking issue created (label: `rfc/proposed`)
- [ ] RFC pull request submitted and ready for discussion (label: `rfc/exploring`)
- [ ] Last call for comments (labels: `rfc/exploring` and `rfc/final-comments`)
- [ ] Accepted and merged RFC pull request (label: `rfc/accepted-future`)
- [ ] Green-light for inclusion in a draft specification, and the author is creating and iterating on pull requests (label: `rfc/accepted-draft`)
- [ ] Pull requests are merged in to a draft specification (label: `rfc/accepted-staged`)
- [ ] Officially published in a non-draft revision of the specification (label: `rfc/released`)

Please close this tracking issue when the proposal enters the `Released` stage of the process.

## Open Points

None yet
---

> The author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue.

PR #126, opened 2026-04-13
feat(rfc): JSON embedded files

crowecawcaw:
**Tracking Issue**: https://github.com/OpenJobDescription/openjd-specifications/issues/124

This is a request for comments about JSON embedded files.

---

### DIFF
--- a/rfcs/0008-json-embedded-files.md
+++ b/rfcs/0008-json-embedded-files.md
@@ -1,0 +1,309 @@
+* Feature Name: json_embedded_files
+* Author(s): Stephen Crowe
+* RFC Tracking Issue: https://github.com/OpenJobDescription/openjd-specifications/issues/124
+* Start Date: 2026-04-07
+* Specification Version: 2023-09 extension EXPR
+* Accepted On: (pending)
+
+## Summary
+
+This RFC adds a new `type: "JSON"` variant of embedded files where `data` is a structured
+object rather than raw text. The runtime resolves format string expressions at the value level
+within the parsed data structure, then serializes the result to JSON when writing the file.
+This eliminates an entire class of quoting and escaping bugs that arise when parameter values
+containing special characters (apostrophes, quotes, colons, etc.) are substituted into
+hand-written YAML/JSON text blobs.
+
+## Basic Examples
+
+### Current approach (broken by special characters)
+
+```yaml
+embeddedFiles:
+  - name: initData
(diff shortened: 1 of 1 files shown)


rfcs/0008-json-embedded-files.md
```diff
@@ -0,0 +1,309 @@
...
+
+### The quoting problem
+
+`type: TEXT` embedded files are the primary way Deadline Cloud integrations pass structured
+configuration data (init-data, run-data) to adaptors. Template authors write YAML or JSON as
+raw text with `{{Param.*}}` placeholders, and the runtime performs text substitution *after*
+the template has been parsed. If the substituted value contains characters that break the
+quoting chosen at serialization time, the worker gets a parse error.
+
+This issue has caused real bugs:
```
leongdl, line 103: Totally agree this is a big problem!

rfcs/0008-json-embedded-files.md
```diff
@@ -0,0 +1,309 @@
...
+
+```yaml
+embeddedFiles:
+  - name: renderConfig
+    filename: config.json
+    type: JSON
+    data:
+      renderer: "{{Param.Renderer}}"
+      resolution: ["{{Param.Width}}", "{{Param.Height}}"]
+      passes:
```
leongdl, line 86: Can we define a limit to the max depth? Say 2?

rfcs/0008-json-embedded-files.md
```diff
@@ -0,0 +1,309 @@
...
+```
+
+The runtime parses `data` as a structured object, resolves each format string expression
+within the values, then serializes the entire structure to JSON. The JSON serializer handles
+all quoting and escaping correctly because substitution happens on the parsed data structure,
+not inside a serialized text blob.
+
+The written file `init-data.json` would contain:
+
+```json
```
leongdl, line 63: I'd like to know or think about if we add limits. EG: Max number of keys, max length of keys and values. 

Thinking about pen test overflow like cases. If we define something reasonable it would be constraint.

rfcs/0008-json-embedded-files.md
crowecawcaw: Do you think there's a need to? The job template itself (YAML or JSON) will be fully parsed before this, and job templates have a max length defined already, so I can't think of a reason we'd need to further limit the JSON blob structure or size.

PR #129, opened 2026-04-16
fix: mark userInterface control as @optional, add conformance tests

### LINKED ISSUE #1109: Bug: Extraneous OpenJD Validation
### Describe the bug

The submitter fails on OpenJD validations that seem extraneous. For example, when you go to submit a job that defines "userInterface" for a parameter, but leave out the "control" field, the submitter will show the user interface fine, but the submission will fail, saying the "control" field is required. But it's very apparent that it should not be required since the UI had no issues drawing itself and a UI is not even needed for submission anyway.

This screenshot shows a job with a integer parameter. The spin box for the integer parameter shows up, but a validation error prevents the submission.
Image

### Expected Behaviour

The submitter should do all UI validations on open and it should relax its validation so that minor errors like leaving out "control" doesn't cause any issues.

### Current Behaviour

It unnecessarily stops the submission.

### Reproduction Steps

Run `deadline bundle gui-submit` on the following template:

[template.yaml](https://github.com/user-attachments/files/26793172/template.yaml)

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.)
2. Output of `python3 --version`
3. Output of `deadline --version`
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer?

Please share other details about your environment that you think might be relevant to reproducing the bug.

crowecawcaw:
Add the @optional annotation to the control field within userInterface for all parameter types (§2.1-2.4, §2.9-2.16). The prose already describes defaults for each type when control is not provided; this change makes the schema consistent with that prose.

Add conformance tests verifying that templates with userInterface containing only label or groupLabel (without control) are valid.

This change is minor and backwards compatible, so recommend merging it directly instead of going through an RFC. The prose already described `control` as optional so this change is more of a bug fix than a feature.

Fixes: https://github.com/aws-deadline/deadline-cloud/issues/1109

---

### DIFF
--- a/conformance-tests/2023-09/base/job_templates/2.1--string-ui-grouplabel-without-control.yaml
+++ b/conformance-tests/2023-09/base/job_templates/2.1--string-ui-grouplabel-without-control.yaml
@@ -1,0 +1,15 @@
+specificationVersion: jobtemplate-2023-09
+name: TestJob
+parameterDefinitions:
+- name: MyParam
+  type: STRING
+  userInterface:
+    groupLabel: My Group
+steps:
+- name: Step1
+  script:
+    actions:
+      onRun:
+        command: echo
+        args:
+        - "{{Param.MyParam}}"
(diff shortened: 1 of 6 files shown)


crowecawcaw:
Updated to clarify intent of `userInterface` section as type hints for which it's reasonable to ignore irrelevant hints.

mwiebe (approved):

jericht (approved):

PR #130, opened 2026-04-17
rfc: Environment Wrap Actions

leongdl:
## PR for RFC

**Tracking Issue**: https://github.com/OpenJobDescription/openjd-specifications/issues/132

This is a request for comments about **RFC 0008 — Environment Wrap Actions**, which
extends `<Environment>` with three new session actions — `onWrapEnter`,
`onWrapTaskRun`, and `onWrapExit` — that let an environment template intercept and
wrap the lifecycle actions of *inner* environments and tasks. A companion opt-out,
`runOnHost: true` on `<Action>`, lets individual actions bypass wrapping when they
must run on the host (credential fetching, mount setup, cleanup that must always
run).

The primary motivation is portable container support: a Docker or Apptainer
environment template can start a container in `onEnter`, route every inner
environment's `onEnter`/`onExit` and every task's `onRun` into the container via the
three wrap hooks, and stop the container in `onExit`. Job templates and inner
environments remain portable across Conda, Rez, Docker, and Apptainer. The design
also generalizes to remote execution, session-wide instrumentation, and privilege
isolation.

This RFC is gated by the new `WRAP_ACTIONS` extension and depends on RFC 0002
(Model Extensions), RFC 0005 (Expression Language), and RFC 0006 (Expression
Function Library — for `repr_sh()` and friends).

---

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.

### DIFF
--- a/conformance-tests/2023-09/WRAP_ACTIONS/env_templates/4.2--empty-actions-no-wrap-no-enter-no-exit.invalid.yaml
+++ b/conformance-tests/2023-09/WRAP_ACTIONS/env_templates/4.2--empty-actions-no-wrap-no-enter-no-exit.invalid.yaml
@@ -1,0 +1,10 @@
+# An environment whose `actions` block is empty is invalid: at least one of
+# onEnter, onWrapEnvEnter, onWrapTaskRun, onWrapEnvExit, or onExit must be defined.
+specificationVersion: environment-2023-09
+extensions:
+- WRAP_ACTIONS
+- EXPR
+environment:
+  name: EmptyActions
+  script:
+    actions: {}
--- a/conformance-tests/2023-09/WRAP_ACTIONS/env_templates/4.2--minimal-wrap-env-template.yaml
+++ b/conformance-tests/2023-09/WRAP_ACTIONS/env_templates/4.2--minimal-wrap-env-template.yaml
@@ -1,0 +1,23 @@
+specificationVersion: environment-2023-09
+extensions:
+- WRAP_ACTIONS
+- EXPR
+environment:
+  name: MinimalWrapEnv
+  script:
+    actions:
+      onEnter:
+        command: echo
+        args: ["entered"]
+      onWrapEnvEnter:
+        command: echo
+        args: ["wrap-enter", "{{WrappedEnv.Name}}"]
+      onWrapTaskRun:
+        command: echo
+        args: ["wrap-task", "{{WrappedAction.Command}}"]
+      onWrapEnvExit:
+        command: echo
+        args: ["wrap-exit", "{{WrappedEnv.Name}}"]
+      onExit:
+        command: echo
+        args: ["exited"]
--- a/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-empty-environment-list.test.yaml
+++ b/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-empty-environment-list.test.yaml
@@ -1,0 +1,47 @@
+# RFC 0008: WrappedAction.Environment is an empty list when no openjd_env
+# was emitted in the session. Verifies the empty case doesn't trip the
+# wrap script's iteration logic.
+template:
+  specificationVersion: jobtemplate-2023-09
+  name: WrappedActionEnvironmentEmpty
+  steps:
+  - name: Step1
+    script:
+      actions:
+        onRun:
+          command: echo
+          args: ["placeholder"]
+
+environments:
+- specificationVersion: environment-2023-09
+  extensions:
+  - WRAP_ACTIONS
+  - EXPR
+  environment:
+    name: WrapEnv
+    script:
+      actions:
+        onWrapEnvEnter:
+          command: python
+          args:
+          - "-c"
+          - "import subprocess,sys;sys.exit(subprocess.run([{{repr_py(WrappedAction.Command)}}]+{{repr_py(WrappedAction.Args)}}).returncode)"
+        onWrapTaskRun:
+          command: python
+          args:
+          - "-c"
+          # Print env count + a sentinel so we can confirm empty-list
+          # handling. With no openjd_env emissions, the loop body should
+          # not execute at all and only the sentinel appears.
+          - "for e in {{repr_py(WrappedAction.Environment)}}:\n print(f'ENV_PRESENT={e}')\nprint('ENV_DUMP_COMPLETE')"
+        onWrapEnvExit:
+          command: python
+          args:
+          - "-c"
+          - "import subprocess,sys;sys.exit(subprocess.run([{{repr_py(WrappedAction.Command)}}]+{{repr_py(WrappedAction.Args)}}).returncode)"
+
+expected:
+  output:
+  - ENV_DUMP_COMPLETE
+  forbidden:
+  - ENV_PRESENT=
--- a/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-enter-and-exit-three-hooks.test.yaml
+++ b/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-enter-and-exit-three-hooks.test.yaml
@@ -1,0 +1,61 @@
+# RFC 0008: a single wrap environment declares all three hooks
+# (onWrapEnvEnter, onWrapTaskRun, onWrapEnvExit). This test validates that
+# all three intercept correctly in one session, with the inner env's
+# onEnter/onExit replaced by the wrap scripts.
+template:
+  specificationVersion: jobtemplate-2023-09
+  name: ThreeHooks
+  jobEnvironments:
+  - name: InnerEnv
+    script:
+      actions:
+        onEnter:
+          command: echo
+          args: ["INNER_ENTER_NORMAL"]
+        onExit:
+          command: echo
+          args: ["INNER_EXIT_NORMAL"]
+  steps:
+  - name: Step1
+    script:
+      actions:
+        onRun:
+          command: echo
+          args: ["ORIGINAL_TASK"]
+
+environments:
+- specificationVersion: environment-2023-09
+  extensions:
+  - WRAP_ACTIONS
+  - EXPR
+  environment:
+    name: WrapEnv
+    script:
+      actions:
+        onWrapEnvEnter:
+          command: python
+          args:
+          - "-c"
+          - "print('WRAP_ENTER_NAME={{WrappedEnv.Name}}');print('WRAP_ENTER_CMD={{WrappedAction.Command}}')"
+        onWrapTaskRun:
+          command: python
+          args:
+          - "-c"
+          - "print('WRAP_TASK_CMD={{WrappedAction.Command}}');print('WRAP_TASK_STEP={{WrappedStep.Name}}')"
+        onWrapEnvExit:
+          command: python
+          args:
+          - "-c"
+          - "print('WRAP_EXIT_NAME={{WrappedEnv.Name}}')"
+
+expected:
+  output:
+  - WRAP_ENTER_NAME=InnerEnv
+  - WRAP_ENTER_CMD=echo
+  - WRAP_TASK_CMD=echo
+  - WRAP_TASK_STEP=Step1
+  - WRAP_EXIT_NAME=InnerEnv
+  forbidden:
+  - INNER_ENTER_NORMAL
+  - INNER_EXIT_NORMAL
+  - ORIGINAL_TASK
--- a/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-exit-status-python.test.yaml
+++ b/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-exit-status-python.test.yaml
@@ -1,0 +1,42 @@
+# RFC 0008: Cross-platform (Python) variant of exit-status propagation.
+# The wrapped action exits non-zero; the wrap script must propagate it.
+# Runs on all platforms.
+template:
+  specificationVersion: jobtemplate-2023-09
+  name: ExitStatusPython
+  steps:
+  - name: Step1
+    script:
+      actions:
+        onRun:
+          command: python
+          args: ["-c", "import sys; sys.exit(42)"]
+
+environments:
+- specificationVersion: environment-2023-09
+  extensions:
+  - WRAP_ACTIONS
+  - EXPR
+  environment:
+    name: WrapEnv
+    script:
+      actions:
+        onWrapEnvEnter:
+          command: python
+          args:
+          - "-c"
+          - "import subprocess,sys; sys.exit(subprocess.run([{{repr_py(WrappedAction.Command)}}]+{{repr_py(WrappedAction.Args)}}).returncode)"
+        onWrapTaskRun:
+          command: python
+          args:
+          - "-c"
+          - "import subprocess,sys; sys.exit(subprocess.run([{{repr_py(WrappedAction.Command)}}]+{{repr_py(WrappedAction.Args)}}).returncode)"
+        onWrapEnvExit:
+          command: python
+          args:
+          - "-c"
+          - "import subprocess,sys; sys.exit(subprocess.run([{{repr_py(WrappedAction.Command)}}]+{{repr_py(WrappedAction.Args)}}).returncode)"
+
+expected:
+  taskFailure:
+    exitCode: 42
--- a/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-exit-status-windows.test.yaml
+++ b/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-exit-status-windows.test.yaml
@@ -1,0 +1,44 @@
+# RFC 0008: Windows variant of exit-status propagation.
+# The wrapped action exits non-zero; the wrap script must propagate it.
+runOn:
+- windows
+
+template:
+  specificationVersion: jobtemplate-2023-09
+  name: ExitStatusWindows
+  steps:
+  - name: Step1
+    script:
+      actions:
+        onRun:
+          command: cmd
+          args: ["/c", "exit /b 42"]
+
+environments:
+- specificationVersion: environment-2023-09
+  extensions:
+  - WRAP_ACTIONS
+  - EXPR
+  environment:
+    name: WrapEnv
+    script:
+      actions:
+        onWrapEnvEnter:
+          command: cmd
+          args:
+          - "/c"
+          - "{{ repr_cmd(WrappedAction.Command) }} {{ repr_cmd(WrappedAction.Args) }}"
+        onWrapTaskRun:
+          command: cmd
+          args:
+          - "/c"
+          - "{{ repr_cmd(WrappedAction.Command) }} {{ repr_cmd(WrappedAction.Args) }}"
+        onWrapEnvExit:
+          command: cmd
+          args:
+          - "/c"
+          - "{{ repr_cmd(WrappedAction.Command) }} {{ repr_cmd(WrappedAction.Args) }}"
+
+expected:
+  taskFailure:
+    exitCode: 42
--- a/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-glob-characters-literal.test.yaml
+++ b/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-glob-characters-literal.test.yaml
@@ -1,0 +1,50 @@
+# RFC 0008 "Recommended test case" #4 — shell globbing.
+# Glob characters (*, ?, [abc]) must reach the wrap script unexpanded.
+# repr_sh() quotes them so the shell does not expand them at wrap time.
+template:
+  specificationVersion: jobtemplate-2023-09
+  name: GlobCharactersLiteral
+  steps:
+  - name: Step1
+    script:
+      actions:
+        onRun:
+          command: ls
+          args:
+          - "*.txt"
+          - "?.log"
+          - "[abc]*"
+
+environments:
+- specificationVersion: environment-2023-09
+  extensions:
+  - WRAP_ACTIONS
+  - EXPR
+  environment:
+    name: WrapEnv
+    script:
+      actions:
+        onWrapEnvEnter:
+          command: bash
+          args:
+          - "-c"
+          - "{{ repr_sh(WrappedAction.Command) }} {{ repr_sh(WrappedAction.Args) }}"
+        onWrapTaskRun:
+          command: bash
+          args:
+          - "-c"
+          - "printf 'ARG=%s\\n' {{ repr_sh(WrappedAction.Args) }}"
+        onWrapEnvExit:
+          command: bash
+          args:
+          - "-c"
+          - "{{ repr_sh(WrappedAction.Command) }} {{ repr_sh(WrappedAction.Args) }}"
+
+runOn:
+- posix
+
+expected:
+  output:
+  - "ARG=*.txt"
+  - "ARG=?.log"
+  - "ARG=[abc]*"
--- a/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-ignores-wrapped-action-vars.test.yaml
+++ b/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-ignores-wrapped-action-vars.test.yaml
@@ -1,0 +1,45 @@
+# The wrap action is not required to reference WrappedAction.* at all — it
+# can run its own logic. This test verifies the wrap runs successfully
+# even when it ignores every injected variable.
+template:
+  specificationVersion: jobtemplate-2023-09
+  name: WrapIgnoresWrappedActionSymbols
+  steps:
+  - name: Step1
+    script:
+      actions:
+        onRun:
+          command: echo
+          args: ["ORIGINAL"]
+
+environments:
+- specificationVersion: environment-2023-09
+  extensions:
+  - WRAP_ACTIONS
+  - EXPR
+  environment:
+    name: WrapEnv
+    script:
+      actions:
+        onWrapEnvEnter:
+          command: bash
+          args:
+          - "-c"
+          - "{{ repr_sh(WrappedAction.Command) }} {{ repr_sh(WrappedAction.Args) }}"
+        onWrapTaskRun:
+          command: echo
+          args: ["SELF_CONTAINED"]
+        onWrapEnvExit:
+          command: bash
+          args:
+          - "-c"
+          - "{{ repr_sh(WrappedAction.Command) }} {{ repr_sh(WrappedAction.Args) }}"
+
+runOn:
+- posix
+
+expected:
+  output:
+  - SELF_CONTAINED
+  forbidden:
+  - ORIGINAL
--- a/rfcs/0008-environment-wrap-actions.md
+++ b/rfcs/0008-environment-wrap-actions.md
@@ -1,0 +1,759 @@
+---
+
+* Feature Name: Environment Wrap Actions
+* Author(s): David Leong <[leongdl](https://github.com/leongdl)>
+* RFC Tracking Issue: https://github.com/OpenJobDescription/openjd-specifications/issues/132
+* Start Date: 2026-04-16
+* Specification Version: 2023-09 extension WRAP_ACTIONS
+* Accepted On: (pending)
+* Depends On:
+  * RFC 0002, Model Extensions (https://github.com/OpenJobDescription/openjd-specifications/issues/57)
+  * RFC 0005, Expression Language (https://github.com/OpenJobDescription/openjd-specifications/pull/93)
+  * RFC 0006, Expression Function Library (https://github.com/OpenJobDescription/openjd-specifications/pull/104)
+
+## Summary
+
+This RFC proposes extending `<Environment>` with three new session actions
+(`onWrapEnvEnter`, `onWrapTaskRun`, and `onWrapEnvExit`) that let an environment
+template intercept and wrap the lifecycle actions of *inner* environments
+and tasks. The
+runtime supplies each wrap action with the wrapped action's fields (command, args,
+timeout, environment) via template variables.
+
+The RFC defines a general-purpose wrapping mechanism. Container execution is the
+motivating use case and the focus of the examples, but the same mechanic supports
+remote execution, session-wide instrumentation, privilege shifts, and any other form
+of redirecting *how* an action runs without modifying the action itself.
+
+## Basic Example
+
+This environment template runs inner environments and tasks inside a Docker
+container. The container starts once in `onEnter`, every inner action is forwarded
+into the container via the three wrap hooks, and the container stops in `onExit`.
+The example is intentionally short. A production-ready version with registry
+authentication, image pull policies, and bind-mount parity, along with
+Apptainer, dry-run, SSH, and profiling variants, is checked in as conformance-test
+fixtures in future examples.
+
+```yaml
+specificationVersion: "environment-2023-09"
+extensions:
+  - WRAP_ACTIONS
+  - EXPR
+parameterDefinitions:
+  - name: ContainerImage
+    type: STRING
+    default: "ubuntu:latest"
+
+environment:
+  name: Docker
+  script:
+    actions:
+      onEnter:
+        command: bash
+        args: ["{{Env.File.Enter}}"]
+      onWrapEnvEnter:
+        command: bash
+        args: ["{{Env.File.Wrap}}"]
+        timeout: "{{WrappedAction.Timeout}}"
+      onWrapTaskRun:
+        command: bash
+        args: ["{{Env.File.Wrap}}"]
+        timeout: "{{WrappedAction.Timeout}}"
+      onWrapEnvExit:
+        command: bash
+        args: ["{{Env.File.Wrap}}"]
+        timeout: "{{WrappedAction.Timeout}}"
+      onExit:
+        command: bash
+        args: ["{{Env.File.Exit}}"]
+
+    embeddedFiles:
+      - name: Enter
+        filename: enter.sh
+        type: TEXT
+        data: |
+          #!/usr/bin/env bash
+          set -euo pipefail
+          DOCKER_CONTAINER_ID=$(docker container run --rm --detach \
+              --mount {{ repr_sh("type=bind,src=" + Session.WorkingDirectory + ",dst=" + Session.WorkingDirectory) }} \
+              {{ repr_sh(Param.ContainerImage) }} \
+              bash -c 'sleep infinity')
+          echo "openjd_env: DOCKER_CONTAINER_ID=$DOCKER_CONTAINER_ID"
+
+      # Shared wrap script for all three hooks: repr_sh() produces safely
+      # quoted argv tokens for every wrapped field.
+      - name: Wrap
+        filename: wrap.sh
+        type: TEXT
+        data: |
+          #!/usr/bin/env bash
+          set -euo pipefail
+          docker container exec \
+              "$DOCKER_CONTAINER_ID" \
+              {{ repr_sh(flatten([['-e', e] for e in WrappedAction.Environment])) }} \
+              {{ repr_sh(WrappedAction.Command) }} \
+              {{ repr_sh(WrappedAction.Args) }}
+
+      - name: Exit
+        filename: exit.sh
+        type: TEXT
+        data: |
+          #!/usr/bin/env bash
+          set -euo pipefail
+          docker container stop --timeout 30 "$DOCKER_CONTAINER_ID"
+```
+
+The job template that runs under this wrapping environment is unchanged from one
+that runs without wrapping. See the existing
+[samples repository](https://github.com/OpenJobDescription/openjd-specifications/tree/mainline/samples/job_templates)
+for portable job templates.
+
+### Execution order with a wrapping environment
+
+For a session with a wrapping Docker environment `A` and a non-wrapping inner
+environment `B`, execution proceeds:
+
+```
+A.onEnter                            (wrapping env's own onEnter is never wrapped)
+  B.onEnter                          (via A.onWrapEnvEnter)
+    Task 1: onRun                    (via A.onWrapTaskRun)
+    Task 2: onRun                    (via A.onWrapTaskRun)
+  B.onExit                           (via A.onWrapEnvExit)
+A.onExit                             (wrapping env's own onExit is never wrapped)
+```
+
+The wrapping environment is not required to be the outermost. If `B` were the
+wrapping environment instead, `A` would run normally and `B` would wrap its
+own inner actions and tasks.
+
+When no wrapping environment is present, every action runs normally. The job
+template and the inner environments are unchanged.
+
+## Motivation
+
+Environment templates today can prepare the context in which a job runs:
+installing software, setting environment variables, starting background daemons.
+They cannot change *how* the actions within the session are executed. The existing
+[bash-in-docker sample](https://github.com/OpenJobDescription/openjd-specifications/tree/mainline/samples/job_templates/bash-in-docker)
+demonstrates a workaround: the step environment starts a container and the task
+execs a script inside it. This approach requires the job template to be written
+specifically for Docker, and inner step environments cannot install software inside
+the container. Swap the environment from Docker to Conda, and the job template
+breaks.
+
+The [Design Tenets](https://github.com/OpenJobDescription/openjd-specifications/wiki/Design-Tenets)
+call for portability:
+
+Job templates should be portable in a way to run them, unmodified, with either a
+Conda, Rez, Docker, or Apptainer environment template that provides the software
+environment to run in.
+
+The three wrap hooks close this gap. Two composability properties make the mechanic
+work without new coupling:
+
+1. An outer environment template can modify the execution of inner
+   environments that it does not know about. The wrap hooks operate on
+   whatever `<Action>` fields the inner environment supplies.
+2. An inner environment template does not need to know a wrapper exists. The
+   same inner template runs unchanged with or without a wrapping environment.
+
+### Use cases
+
+1. **Container execution.** Apply a Docker or Apptainer container as the outer
+   environment, and let inner environments install plugins, activate Conda
+   environments, or stage dependencies inside it. The job template and the
+   inner environments are unchanged when the wrapping environment is swapped
+   or removed.
+
+2. **Remote execution.** An environment template can SSH into a remote host, or
+   submit actions to a cloud API, via the three wrap hooks. Inner environment setup
+   and task runs then execute on the remote host rather than locally.
+
+3. **Session-wide instrumentation.** Wrap every action with profiling, tracing, or
+   resource-accounting tools without modifying the job template or inner
+   environments. Task-only profiling (which exists via onRun substitution today)
+   misses the setup phases that often contain performance issues.
+
+4. **Privilege shifts.** Run inner actions as a different user or with reduced
+   capabilities by wrapping the command with `sudo -u`, `unshare`, or similar.
+
+### Backward compatibility
+
+This RFC is additive and gated by the `WRAP_ACTIONS` extension name declared under
+RFC 0002. Specifically:
+
+- Schedulers that do not implement `WRAP_ACTIONS` MUST reject templates that list
+  it in `extensions:`, per RFC 0002's extension-handling rules.
+- Schedulers that do implement `WRAP_ACTIONS` MUST ignore the three wrap hooks on
+  templates that do not list `WRAP_ACTIONS` in `extensions:`, so that existing
+  templates continue to behave exactly as before.
+- No existing field changes meaning; `onEnter` and `onExit` continue to behave as
+  they do today when no wrap hook is active.
+
+**`EXPR` is a hard prerequisite of `WRAP_ACTIONS`.** Templates that list
+`WRAP_ACTIONS` MUST also list `EXPR` (RFC 0005). Schedulers MUST reject
+templates that list `WRAP_ACTIONS` without also listing `EXPR`. The
+function library defined in RFC 0006 (including `repr_sh`/`repr_cmd`/
+`repr_pwsh` and `flatten`) is part of the same `EXPR` extension, so no
+separate extension name is required.
+
+The rationale: safe reconstruction of a wrapped command line from
+`WrappedAction.*` requires shell-aware quoting (see
+[Security Considerations](#security-considerations)), and `EXPR` plus its
+function library is the specification-provided mechanism for that.
+
+## Specification
+
+### Terminology
+
+The following terms are used throughout this specification.
+
+| Term                    | Definition                                                                                                                                               |
+|-------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **Host**                | The OS environment of the worker where the OpenJD session runtime runs and launches actions.                                                             |
+| **Session runtime**     | The OpenJD implementation running on the worker that launches actions, scans their stdout for `openjd_*` macros, enforces timeouts, and delivers signals. Distinct from the (cloud-side) scheduler that accepts templates and dispatches sessions. |
+| **Wrap hook**           | One of the three new `<EnvironmentActions>` fields: `onWrapEnvEnter`, `onWrapTaskRun`, `onWrapEnvExit`.                                                        |
+| **Wrap script**         | The program supplied by a wrap hook's `<Action>`, and the process the session runtime launches to execute it. The wrap script itself runs normally; it chooses what to do with the wrapped action's fields. |
+| **Wrapping environment**| The environment that defines wrap hooks. Its own `onEnter` and `onExit` run normally and are never wrapped.                                              |
+| **Inner environment**   | Any environment in the session stack that is inner to the wrapping environment. Its lifecycle actions (`onEnter`, `onExit`) are intercepted by wrap hooks. |
+| **Wrapped action**      | An inner environment's `onEnter`/`onExit`, or a task's `onRun`, whose execution is intercepted by a wrap hook. The wrap script receives the action's fields via `WrappedAction.*` template variables. |
+| **Wrapped process**     | The OS process the wrap script invokes to perform the wrapped action (e.g., `docker container exec ...`, `ssh ...`). A grand-child of the session runtime. |
+
+### Schema modifications
+
+> Changes to [the template schema](https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas).
+
+> A modification to [`<EnvironmentActions>`](https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#43-environmentactions)
+
+```diff
+  <EnvironmentActions> ::= the object:
+    onEnter: <Action>
++   onWrapEnvEnter: <Action>    # @optional @extension WRAP_ACTIONS
++   onWrapTaskRun: <Action>     # @optional @extension WRAP_ACTIONS
++   onWrapEnvExit: <Action>     # @optional @extension WRAP_ACTIONS
+    onExit: <Action> # @optional
+```
+
+1. *onEnter*: the action to run when entering the environment.
+2. *onWrapEnvEnter*: if provided, runs instead of the `onEnter` action of every
+   *inner* environment that enters the session while this environment is
+   active.
+3. *onWrapTaskRun*: if provided, runs instead of the task's `onRun` action
+   for every task that runs while this environment is active.
+4. *onWrapEnvExit*: if provided, runs instead of the `onExit` action of every
+   *inner* environment that exits while this environment is active.
+5. *onExit*: the action to run when exiting the environment.
+
+Each wrap hook receives the wrapped action's context via the
+`WrappedAction.*` template variables (see
+[Template variables](#template-variables)).
+
+**All-or-nothing rule.** An environment that defines any of `onWrapEnvEnter`,
+`onWrapTaskRun`, or `onWrapEnvExit` MUST define all three, ensuring execution
+context parity across every wrapped lifecycle phase. Schedulers MUST reject
+templates that define only a subset of the wrap hooks before the session
+begins.
+
+**No changes to `<Action>`.** This RFC does *not* modify the `<Action>` schema. All
+changes are additive at the `<EnvironmentActions>` level, and the new template
+variables are read-only context supplied to wrap actions by the runtime.
+
+### Wrap ordering with multiple environments
+
+A session MUST have at most one wrapping environment. Schedulers MUST reject
+sessions whose stack contains two or more environments that define any wrap
+hook, before entering any environment.
+
+A wrapping environment's wrap hooks intercept only the actions of inner
+environments and tasks. The wrapping environment's own `onEnter` and `onExit`
+are never wrapped.
+
+### Template variables
+
+The following variables are read-only context supplied by the runtime to wrap
+actions. They have identical names and semantics across all three hooks, so
+helper scripts can be reused unchanged.
+
+**Available in `onWrapEnvEnter`, `onWrapTaskRun`, and `onWrapEnvExit`:**
+
+| Variable                          | Type           | Description |
+|-----------------------------------|----------------|-------------|
+| `WrappedAction.Command`           | `string`       | The `command` from the wrapped action. |
+| `WrappedAction.Args`              | `list[string]` | The `args` from the wrapped action. |
+| `WrappedAction.Environment`       | `list[string]` | Environment variables defined by `openjd_env` earlier in the session, as `["KEY=value", ...]`. See [Host environment variables and embedded file paths](#host-environment-variables-and-embedded-file-paths). |
+| `WrappedAction.Timeout`           | `int`          | The timeout in seconds specified on the wrapped action, or `0` if none. See [Timeout behavior](#timeout-behavior). |
+
+**Additionally available in `onWrapEnvEnter` and `onWrapEnvExit`:**
+
+| Variable                          | Type           | Description |
+|-----------------------------------|----------------|-------------|
+| `WrappedEnv.Name`                 | `string`       | The `name` of the inner environment whose action is being wrapped. |
+
+**Additionally available in `onWrapTaskRun`:**
+
+| Variable                          | Type           | Description |
+|-----------------------------------|----------------|-------------|
+| `WrappedStep.Name`                | `string`       | The `name` of the step whose task is being wrapped. |
+
+Templates MUST NOT reference `WrappedAction.*` outside the three wrap hooks,
+`WrappedEnv.*` outside `onWrapEnvEnter` and `onWrapEnvExit`, or `WrappedStep.*`
+outside `onWrapTaskRun`. Schedulers MUST reject templates that violate
+this scope rule.
+
+### Host environment variables and embedded file paths
+
+`WrappedAction.Environment` carries only `openjd_env`-defined variables from
+the current session. Host-inherited variables (`HOME`, `PATH`, `OPENJD_*`,
+etc.) and host filesystem paths referenced in `WrappedAction.Command` or
+`WrappedAction.Args` (for example, `{{Env.File.X}}` resolves to a host path)
+are not surfaced and are the wrap environment's responsibility to forward or
+path-map into the wrapped execution context. Failure manifests as
+`command not found` or `No such file or directory` from the wrapped
+process.
+
+### Forwarding `<Action>` fields, current and future
+
+The `WrappedAction.*` namespace is the single, standardized surface for
+every field of the wrapped `<Action>`. The fields listed in
+[Template variables](#template-variables) (`Command`, `Args`, `Environment`,
+`Timeout`) cover today's `<Action>` schema. When a future specification
+adds a field to `<Action>` — for example, the `logMessageTimeout` idea from
+[Discussion #118](https://github.com/OpenJobDescription/openjd-specifications/discussions/118)
+— it MUST surface under the same namespace using the field's PascalCased
+name (e.g., `WrappedAction.LogMessageTimeout`) without modification to this
+RFC. New fields gated by an `@extension` activate under this namespace
+exactly when that extension is active in the template.
+
+This keeps the contract between the runtime and wrap scripts stable across
+schema evolution: a wrap script written today references existing
+`WrappedAction.*` fields; a wrap script written for a later spec version
+references the same namespace, with new entries available when the
+corresponding extension is in use.
+
+Wrap scripts MAY reference any field they recognize and MUST tolerate
+fields they do not. A wrapped action that relies on a field whose wrap
+script does not propagate it behaves as if the field were absent.
+
+### Modifications to How Jobs Are Run
+
+> Modifications to [How Jobs Are Run](https://github.com/OpenJobDescription/openjd-specifications/wiki/How-Jobs-Are-Run)
+
+```diff
+  Once the Environments have been created and entered for a Session, a series of Tasks are
+  run within that Session and the Environments. Tasks from any Step in a Job can run within
+  the same Session provided that the set of Environments that are required to run those
+  Tasks are identical.
+
++ If exactly one Environment in the session stack defines the wrap hooks (onWrapEnvEnter,
++ onWrapTaskRun, onWrapEnvExit), then the lifecycle actions of inner Environments and
++ tasks are intercepted:
++
++ - An inner Environment's onEnter is replaced by the wrapping Environment's onWrapEnvEnter.
++ - A task's onRun is replaced by the wrapping Environment's onWrapTaskRun.
++ - An inner Environment's onExit is replaced by the wrapping Environment's onWrapEnvExit.
++
++ The wrapping Environment's own onEnter and onExit are never wrapped; they always run
++ normally. If more than one Environment in the session stack defines any wrap hook,
++ the session is invalid and the scheduler must reject it before entering any
++ Environment. If an environment defines any wrap hook, it must define all three.
+```
+
+### Stdout forwarding and macro propagation
+
+- Wrap scripts MUST forward the wrapped process's stdout and stderr to their
+  own stdout and stderr verbatim, without buffering, filtering, or
+  transformation. `docker container exec`, `apptainer exec`, and `ssh`
+  satisfy this by default; authors of custom wrappers MUST preserve it.
+- OpenJD session runtimes MUST scan the wrap script's stdout for OpenJD stdout macros
+  (`openjd_env`, `openjd_fail`, `openjd_progress`, `openjd_status`, and any
+  future macros) and MUST NOT scan the grand-child's stdout directly. A
+  macro emitted by a wrapped process is recognized identically to one
+  emitted by the wrap script itself.
+- OpenJD session runtimes MUST include in `WrappedAction.Environment` every
+  `openjd_env`-defined variable emitted by any earlier action in the same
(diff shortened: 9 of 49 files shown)


rfcs/0008-environment-wrap-task-run.md
```diff
@@ -0,0 +1,1342 @@
...
+  * RFC 0006 — Expression Function Library (https://github.com/OpenJobDescription/openjd-specifications/pull/104)
+
+## Summary
+
+This RFC proposes extending `<Environment>` with three new session actions —
+`onWrapEnter`, `onWrapTaskRun`, and `onWrapExit` — that let an environment template
+intercept and wrap the lifecycle actions of *inner* environments and tasks. The runtime
+supplies each wrap action with the wrapped action's command, args, timeout, cancelation
+method, and environment variables as template variables. A companion opt-out,
+`runOnHost: true` on `<Action>`, lets individual actions bypass wrapping when they
```
mwiebe: Could we make this `runOnHost` also be controlled by the wrapping environment? Then an author of an external environment could have more control over this. If we leave this as is we might want to find a different name, because the distinguishing factor is that it skips the wrapping - both wrapped and non-wrapped actions are running on the host.

rfcs/0008-environment-wrap-task-run.md
```diff
@@ -0,0 +1,1342 @@
...
+must run on the host (credential fetching, mount setup, cleanup that must always run).
+
+The primary motivation is container support: a Docker or Apptainer environment template
```
mwiebe: This is the primary motivation, but it's worth mentioning that we want this to be a general composable feature to the extent possible. If we can avoid all references to containers in the structure of the extension (vs the docs and examples which should have many), I think that will be useful.

rfcs/0008-environment-wrap-task-run.md
```diff
@@ -0,0 +1,1342 @@
...
+        command: "bash"
+        args: ["{{Env.File.Enter}}"]
+      onWrapEnter:
+        command: "bash"
+        args: ["{{Env.File.WrapEnter}}"]
+        timeout: "{{Env.Wrapped.Timeout}}"
+      onWrapTaskRun:
+        command: "bash"
+        args: ["{{Env.File.WrapTaskRun}}"]
+        timeout: "{{Task.Timeout}}"
```
mwiebe: The difference between `Env.Wrapped.Timeout` and `Task.Timeout` feels unnecessary. Switching to either `Task.Wrapped.Timeout`, or `Env.Timeout` would make it consistent. Since the timeout is on the action itself, maybe it could be `Task.Action.Timeout` and `Env.Action.Timeout`?

We should also think about the general forwarding mechanism. How will we forward https://github.com/OpenJobDescription/openjd-specifications/discussions/118 when we make that? Can there be a general mechanism, or will we give the template author responsibility to plumb through the features?

rfcs/0008-environment-wrap-task-run.md
```diff
@@ -0,0 +1,1342 @@
...
+        # before SIGKILL.
+        docker container stop \
+            $DOCKER_CONTAINER_ID \
+            --timeout 30
+```
+
+### Apptainer environment template
+
+The same pattern applies to Apptainer (daemonless, each wrap hook invokes
+`apptainer exec` directly rather than exec'ing into a running container). See
```
mwiebe: Can't it run as a daemon? This seems necessary for many use cases.

rfcs/0008-environment-wrap-task-run.md
```diff
@@ -0,0 +1,1342 @@
...
+[Appendix A: Apptainer environment template](#appendix-a-apptainer-environment-template)
+for the full example.
+
+### Job template that works with any environment
```
mwiebe: It should just be able to point to our existing samples. Maybe add a technical requirement that the way we've been writing jobs already must work with or without wrapping to the extent we can do that.

I guess the purpose of this is to show the `runOnHost` option?

rfcs/0008-environment-wrap-task-run.md
```diff
@@ -0,0 +1,1342 @@
...
+execution context for the entire session — including inner environments' lifecycle
+actions — and the job template specifies the work. Neither needs to know about the
+other. The `runOnHost` opt-out gives specific actions a clean way to decline wrapping
+when they fundamentally cannot work inside the wrapped context.
+
+This separation is a proven pattern across workflow ecosystems. Nextflow, CWL, Snakemake,
+and WDL all solve the same problem — abstracting the container runtime from the job logic —
+using language-specific mechanisms (see Prior Art). OpenJD's approach is unique in using
+composable environment templates rather than a global config toggle, which enables per-queue
+and per-step runtime selection in multi-step pipelines.
```
mwiebe: I don't think per-step runtime selection is what we're looking for. This paragraph could focus more on the technical details like what aspects of composability are important for the design, and less on "proven pattern" and "is unique" phrasing.

rfcs/0008-environment-wrap-task-run.md
```diff
@@ -0,0 +1,1342 @@
...
+
+1. **Containers with inner-env setup.** Run a studio-wide Docker or Apptainer container
+   at the queue level, and let per-step environments install plugins, activate Conda
+   environments, or stage dependencies *inside* the container. The job template and
+   the step environments are unchanged when the wrapping environment is swapped or
+   removed.
+
+2. **Host-level actions alongside containers.** Actions that must run on the host —
+   mounting an NFS share the container will bind-mount, setting up a VPN tunnel for a
+   license server, fetching short-lived credentials — use `runOnHost: true` to bypass
```
mwiebe: These are environment setup actions, not workload properties. They seem to belong in environments that could be nested outside the docker one?

rfcs/0008-environment-wrap-task-run.md
```diff
@@ -0,0 +1,1342 @@
...
+   actions to a cloud API, using the same three wrap hooks. Inner environment setup runs
+   on the remote host, not locally.
+
+4. **Instrumentation over the full session.** Wrap every action with profiling, tracing,
+   or resource-accounting tools without modifying the job template or the inner
+   environments. Profiling only tasks (current RFC) misses the setup phases that are
+   often where performance problems hide.
+
+5. **Privilege isolation.** Run inner actions as a different user or with reduced
+   capabilities by wrapping the command with `sudo -u`, `unshare`, or a jailed shell.
```
mwiebe: If the inner environments and task runs can select `runOnHost` then this use case is weaker.

rfcs/0008-environment-wrap-task-run.md
```diff
@@ -0,0 +1,1342 @@
...
+   wrap hooks. See [Rejected Ideas › Nested wrap composition](#nested-wrap-composition).
+2. **Periodic health-checking of wrapped processes** (e.g., detecting stalled
+   containers). Deferred to a separate RFC for Environment monitoring.
+3. **Wrap-aware cancelation.** The wrap hook's own `<Cancelation>` is used for any
+   wrapped action; the wrapped action's `<Cancelation>` is not surfaced to the wrap
+   script. See [Cancelation behavior](#cancelation-behavior).
+4. **Cross-OS wrapping.** The same-path bind-mount requirement assumes the host and
+   the wrapped execution context share path-separator conventions (Linux host with
+   Linux container, or Windows with Windows). Cross-OS wrapping (e.g., a Windows host
+   launching a Linux container) is not supported by this RFC.
```
mwiebe: The responsibility of the RFC would be to provide all information necessary to do this Cross-OS wrapping, not to assist in any way. I think it does that?

rfcs/0008-environment-wrap-task-run.md
```diff
@@ -0,0 +1,1342 @@
...
+enter, every task, and every exit runs inside it, and the container stops when the
+session ends. State persists across actions within the session (loaded plugins, cached
+data, pip-installed packages), which matches how DCC applications work.
+
+This is the same amortization pattern that task stickiness provides for background
+daemons, and that task chunking (RFC 0001) provides for frame ranges.
+
+### Security: container isolation boundaries
+
+Each container instance should be scoped to a single security boundary. In the
```
mwiebe: The spec isn't about containers, it's about the ability to perform action wrapping reliably. I think the scheduler doesn't have any responsibility about "container," its responsibility is purely about providing all the specified arguments to the wrap* actions. The RFC should focus exclusively on the wrapping part where it is about the OpenJD behavior, and talk about containers in example contexts that are about containers.

rfcs/0008-environment-wrap-task-run.md
```diff
@@ -0,0 +1,1342 @@
...
+cannot be reused by a different queue's session, even on the same worker host.
+
+Schedulers implementing the wrap hooks should ensure that:
+
+1. The container is stopped and removed in `onExit`, even on failure, so no state leaks
+   between sessions.
+2. The container does not run with elevated privileges (`--privileged`) unless explicitly
+   configured by the environment template author.
+3. Bind mounts are scoped to the session working directory and explicitly declared paths,
+   not the entire host filesystem.
```
mwiebe: None of these are scheduler responsibilities - they are the responsibility of the environment implementer writing support for containers. The spec should ensure that the environment has sufficient capabilities to do what it needs to do.

rfcs/0008-environment-wrap-task-run.md
```diff
@@ -0,0 +1,1342 @@
...
+  <EnvironmentActions> ::= the object:
+    onEnter: <Action>
++   onWrapEnter: <Action>    # @optional
++   onWrapTaskRun: <Action>  # @optional
++   onWrapExit: <Action>     # @optional
+    onExit: <Action> # @optional
+```
+
+1. *onEnter* — The action to run when entering the environment.
+2. *onWrapEnter* — If provided, this action is run instead of the `onEnter` action of
```
mwiebe: Consider adding an all or nothing rule. Either all onWrap* must be included, or none of them. This would help against accidentally implementing just part of the wrapping, then getting hard to debug results because not everything is running where expected.

mwiebe (changes requested):

rfcs/0008-environment-wrap-task-run.md
leongdl: Interesting, unless the wrapping environment inspects and has logic for every line in the wrapped action? 

Something like exporting environment variables on the host would be one case that is hard to detect

(And agree on the naming is hard)

rfcs/0008-environment-wrap-task-run.md
leongdl: Sounds good, and makes sense.

rfcs/0008-environment-wrap-task-run.md
leongdl: Interesting, I did not see 118, let me read and incorporate it.

rfcs/0008-environment-wrap-task-run.md
leongdl: I think daemonless was meaning the "apptainer" daemon, like the persistent "docker daemon", not necessarily referring to the container as a daemon. Maybe a framing difference here.

rfcs/0008-environment-wrap-task-run.md
leongdl: Yeah - exactly. The RunOnHost was an escape hatch. 

Lets chat offline how we can offer the escape hatch. Either within the hook, or as an explicit declaration.

rfcs/0008-environment-wrap-task-run.md
leongdl: Sure, will re-phrase.

rfcs/0008-environment-wrap-task-run.md
leongdl: Yeah, these are enviornment setup actions that does not need to be wrapped. I'll find a better home for these sentences.

rfcs/0008-environment-wrap-task-run.md
leongdl: True, the `runOnHost` escape hatch would change this. But nonetheless, examples such as running a container would be able to use the `sudo -u`.

I'll re-frame this as a addendum use case, not necessarily an argument.

rfcs/0008-environment-wrap-task-run.md
leongdl: True, let me delete it.

Along the way I was thinking - why wrap windows containers in linux, and vice versa. Although linux containers on windows absolutely work but is terrible thorugh WSL.

Was originall thinking to cut some scope or defer this family of cases for later.

rfcs/0008-environment-wrap-task-run.md
leongdl: Sounds good - I'll do a pass and move this into my separate design doc instead.

rfcs/0008-environment-wrap-task-run.md
leongdl: True, I think here, AI here thought "deadline" overall as the scheduler, I'll delete these.

rfcs/0008-environment-wrap-task-run.md
leongdl: Good suggestion! I'll add that into the spec.

While the template yaml / json is slightly larger it is much harder to do the wrong thing.

leongdl:
## Update for RFC 0008 reviewers

The RFC has been substantially revised and synthesized into the 2023‑09 spec and wiki. Highlights since the last review:

**Scope expanded from one hook to three.** The proposal now wraps the full environment lifecycle, not just task run:

- `onWrapEnter` — replaces an inner environment's `onEnter`
- `onWrapTaskRun` — replaces a task's `onRun` (the original proposal)
- `onWrapExit` — replaces an inner environment's `onExit`

The file was renamed accordingly: `0008-environment-wrap-task-run.md` → `0008-environment-wrap-actions.md`. The wrapping environment's own `onEnter`/`onExit` are never wrapped — they always run on the host.

**New hard constraints, enforced before session entry:**

1. *All‑or‑nothing* — an environment must define all three wrap hooks or none.
2. *Single‑layer* — at most one wrapping environment per session stack.
3. *EXPR prerequisite* — templates listing `WRAP_ACTIONS` must also list `EXPR`.

**New template‑variable scopes:**

- `Env.Wrapped.*` — available only inside `onWrapEnter` / `onWrapExit`
- `Task.Action.*` — available only inside `onWrapTaskRun`

Schedulers must reject templates that reference these outside their allowed hooks.

**Timeout semantics clarified.** A wrap hook's `timeout` bounds the host‑side script (infrastructure side); the wrapped action's timeout is exposed to the script via `Task.Action.Timeout` / `Env.Wrapped.Action.Timeout` (workload side, `0` when unset). `onWrapExit` defaults to 300s like `onExit`; the other two have no default.

**Exit‑code 125 reserved for infrastructure failure** (Docker daemon unreachable, image pull failed, SSH refused, etc.). Other non‑zero exits are workload failures. Schedulers may use this to drive retry / worker‑eviction / alerting.

**Stdout macro handling specified.** Schedulers scan the wrap script's stdout for OpenJD macros (`openjd_env`, `openjd_unset_env`, …), not the wrapped process's. Wrap scripts must forward the wrapped process's stdout/stderr verbatim, so macros emitted by the wrapped process are recognized identically.

**Synthesized into the normative docs:**

- `wiki/2023-09-Template-Schemas.md` — `WRAP_ACTIONS` registered in the extensions list; hooks, scopes, timeout table, and exit‑125 convention added.
- `wiki/How-Jobs-Are-Run.md` — lifecycle interception described and the stdout‑scanning rule documented.

Re‑review focus suggestions: the all‑or‑nothing + single‑layer rules, the `Env.Wrapped.*` / `Task.Action.*` scope enforcement, and whether `125` is the right sentinel for infrastructure failure.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+      # docker exec pattern: docker handles the kill itself when the inner
+      # process exceeds its time budget, so the wrap script does not need
+      # additional grace beyond the wrapped timeout. Wrappers that perform
+      # cleanup after the wrapped process (SSH teardown, custom remote-exec)
+      # should give the hook timeout extra grace via EXPR arithmetic, e.g.,
+      # timeout: "{{Env.Wrapped.Action.Timeout + 30}}".
+      onWrapEnter:
+        command: bash
+        args: ["{{Env.File.Wrap}}"]
+        timeout: "{{Env.Wrapped.Action.Timeout}}"
```
mwiebe: Thinking some more about the names for the wrapped action variables, these two properties would be useful:

1. They have exactly the same name across all of the `onWrap*` actions.
2. They're compatible with a future introspection where the action can reference itself, probably via `Action.*`.

Based on that, instead of `Env.Wrapped.Action.*` and `Task.Action.*`, the names `WrappedAction.*` is a good candidate to use uniformly.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+
+    embeddedFiles:
+      - name: Enter
+        filename: enter.sh
+        type: TEXT
+        data: |
+          #!/usr/bin/env bash
+          set -euo pipefail
+          DOCKER_CONTAINER_ID=$(docker container run --rm --detach \
+              --mount 'type=bind,src={{Session.WorkingDirectory}},dst={{Session.WorkingDirectory}}' \
```
mwiebe: > 'type=bind,src={{Session.WorkingDirectory}},dst={{Session.WorkingDirectory}}'

For this part we'll want it to be inside of a `repr_sh` instead of in single quotes. Does that need a let binding or is there a different good way to represent it?

mwiebe: I excluded f-strings from the EXPR extension, but maybe that's how it could be more pleasant to write?

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+        data: |
+          #!/usr/bin/env bash
+          set -euo pipefail
+          docker container exec \
+              "$DOCKER_CONTAINER_ID" \
+              {{ repr_sh(flatten([['-e', e] for e in Env.Wrapped.Action.Environment])) }} \
+              {{ repr_sh(Env.Wrapped.Action.Command) }} \
+              {{ repr_sh(Env.Wrapped.Action.Args) }}
+
+      - name: WrapTask
```
mwiebe: Looks like `Wrap` and `WrapTask` could be the same script file if the wrapped action variable names are the same for this example.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+```
+
+The job template that runs under this wrapping environment is unchanged from one
+that runs without wrapping. See the existing
+[samples repository](https://github.com/OpenJobDescription/openjd-specifications/tree/mainline/samples/job_templates)
+for portable job templates.
+
+### Execution order with a wrapping environment
+
+For a session with a wrapping Docker environment `A` and a non-wrapping inner step
```
mwiebe: Is it worth mentioning here that the wrapping environment doesn't have to be outermost? B could be the wrapping environment, then A runs like normal, B wraps the tasks' onRun.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+A.onEnter                            → HOST      (wrapping env's own onEnter is never wrapped)
+  B.onEnter                          → WRAPPED   (via A.onWrapEnter)
+    Task 1: onRun                    → WRAPPED   (via A.onWrapTaskRun)
+    Task 2: onRun                    → WRAPPED   (via A.onWrapTaskRun)
+  B.onExit                           → WRAPPED   (via A.onWrapExit)
+A.onExit                             → HOST
+```
+
+On a queue with no wrapping environment, `A` does not exist and every action runs
+on the host. The job template and the inner environments are unchanged.
```
mwiebe: Instead of "on the host," maybe say "normally" since the wrapped actions typically run on the host as well.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+```
+A.onEnter                            → HOST      (wrapping env's own onEnter is never wrapped)
+  B.onEnter                          → WRAPPED   (via A.onWrapEnter)
+    Task 1: onRun                    → WRAPPED   (via A.onWrapTaskRun)
+    Task 2: onRun                    → WRAPPED   (via A.onWrapTaskRun)
+  B.onExit                           → WRAPPED   (via A.onWrapExit)
+A.onExit                             → HOST
+```
+
+On a queue with no wrapping environment, `A` does not exist and every action runs
```
mwiebe: Open Job Description doesn't define what a queue is. Deadline Cloud queue environments are based on the [environment template specification](https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#12-environment-template). Should reword to ensure only the OpenJD terms are being used.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+specifically for Docker, and inner step environments cannot install software inside
+the container. Swap the environment from Docker to Conda, and the job template
+breaks.
+
+The [Design Tenets](https://github.com/OpenJobDescription/openjd-specifications/wiki/Design-Tenets)
+call for portability:
+
+> Job templates should be portable in a way to run them, unmodified, with either a
+> Conda, Rez, Docker, or Apptainer environment template that provides the software
+> environment to run in.
```
mwiebe: The way this is written it looks like a quote from the tenets, but it's an example application of the portability tenet into this application. Because of that I would remove the `> ` quoting from it to make it clear this text is primary not a quote.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+task render FRAME=123                                              # local
+task render FRAME=123 WRAPPER="docker run -v $(pwd):/data image"   # Docker
+task render FRAME=123 WRAPPER="apptainer exec image.sif"           # Apptainer
+```
+
+Wrap hooks generalize this: the wrap hook has access to the full lifecycle
+context (environment variables, timeout, enter/task/exit distinction) and
+composes through the environment-template model.
+
+### Docker exec pattern in OpenJD
```
mwiebe: Suggest to delete this section, `bash-in-docker` was already referenced above.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+logic.
+
+### Action-scoped template variable namespace
+
+The `Task.Action.*` and `Env.Wrapped.Action.*` naming makes the
+template-variable → `<Action>`-field correspondence explicit: every variable
+under `.Action.` maps to a named field of the wrapped `<Action>` object. This has
+three concrete benefits:
+
+1. Symmetry between the task and environment branches (both use `.Action.`),
```
mwiebe: You can't copy/paste them though, I think matching the full names exactly is better.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+
+### `list[string]` for environment variables
+
+Environment variables are provided as a flat list of `"KEY=value"` strings rather
+than a dictionary. This matches the format expected by container runtimes
+(`docker exec -e KEY=VALUE`, `apptainer exec --env KEY=VALUE`) and is
+straightforward to iterate over in a list comprehension:
+
+```yaml
+{{ repr_sh(flatten([['-e', e] for e in Task.Action.Environment])) }}
```
mwiebe: Can you check if `e.split('=', 1)` works as expected to be able to separately access the key and value?

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+it wraps the value in single quotes and escapes embedded single quotes.
+`repr_sh(list[string])` applies this per element and joins with spaces, so each
+element becomes exactly one argv entry when the shell parses the line.
+
+For Windows shells, the equivalent functions are `repr_cmd()` and `repr_pwsh()`
+from RFC 0006, which handle each shell's different escaping rules.
+
+## Security Considerations
+
+### Command injection
```
mwiebe: I would frame this more that the template author is responsible to ensure the template is correct about quoting all input, and our job in the spec is to provide primitives that enable them to do so. `repr_sh` is key to that.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+### Wrapping overhead and command length
+
+Wrap hooks are subject to the host OS maximum command-line length. A wrap
+hook effectively doubles the command string (once in the wrapper's
+invocation and once in the wrapped action). A job with many `openjd_env`-set
+variables plus a `docker exec` prefix can exceed `ARG_MAX` on a command that
+would fit directly. An `onRun` near the 256 KB macOS `ARG_MAX`, for example,
+may fail only when wrapped.
+
+Implementations SHOULD validate command-line length before exec and produce
```
mwiebe: Who does this affect? It's unclear whether this is about the OpenJD implementation or the template?

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+- **CWL** declares `DockerRequirement` as a hint; runners translate at execution.
+- **Snakemake** declares `container:` per rule and converts at runtime.
+- **WDL** declares `docker:` per task; engines may substitute.
+
+OpenJD's approach uses composable environment templates. The distinction from the
+workflow-DSL prior art is that OpenJD separates the outer execution-context layer
+(the wrap environment) from the job-logic layer (the job template), rather than
+annotating individual processes with a runtime identifier.
+
+### Universal Wrapper pattern
```
mwiebe: This seems like it's describing the idea of this spec, not prior art? I'd suggest to remove this section.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+
+### Docker exec pattern in OpenJD
+
+The existing
+[bash-in-docker sample](https://github.com/OpenJobDescription/openjd-specifications/tree/mainline/samples/job_templates/bash-in-docker)
+demonstrates the pattern of starting a container in a step environment and
+exec'ing tasks into it. This RFC generalizes that pattern to work at any
+environment level and extends it to cover inner-environment setup and teardown.
+
+### Conda and Rez queue environments
```
mwiebe: This section doesn't seem to describe precedent. Both conda and rez have `run` commands that could be used in a wrapping environment though, maybe that's worth mentioning here. Otherwise I'd delete this section.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+### Implicit container support in the runtime
+
+We rejected automatic container detection (e.g., the runtime acts on a
+`containerImage` field) because it couples the specification to specific
+container runtimes and removes the user's ability to customize the
+invocation (mount points, network mode, GPU flags, security options). The
+explicit wrap-hook approach keeps the specification general and the user in
+control.
+
+### Global config toggle for runtime selection
```
mwiebe: Maybe merge this with the above section to tighten up the RFC (We're at line 923! I think it can be shorter).

wiki/2023-09-Template-Schemas.md
```diff
@@ -1550,17 +1558,43 @@ An `<EnvironmentActions>` is the object:
...
+> 1. *All-or-nothing rule.* An environment that defines any of `onWrapEnter`, `onWrapTaskRun`, or
+>    `onWrapExit` must define all three. Schedulers must reject templates that define only a subset of
+>    the wrap hooks before the session begins.
+> 2. *Single-layer rule.* A session must contain at most one environment that defines wrap hooks.
+>    Schedulers must reject sessions whose stack contains two or more such environments before
+>    entering any environment.
+> 3. *EXPR prerequisite.* A template that lists `WRAP_ACTIONS` in `extensions:` must also list `EXPR`.
+>    Schedulers must reject templates that list `WRAP_ACTIONS` without also listing `EXPR`.
+> 4. The wrapping environment's own `onEnter` and `onExit` are never wrapped; they always run on the
+>    host.
```
mwiebe: Suggest "run normally" instead of "run on the host." You'll want to do a sweep for this one throughout.

wiki/2023-09-Template-Schemas.md
```diff
@@ -1618,18 +1652,33 @@ positive integer value in base-10, and:
...
+   (workload-side). When the wrapped action specified no timeout, those variables evaluate to `0`.
+   See [RFC 0008](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0008-environment-wrap-actions.md).
+
 4. *cancelation* — If defined, provides details regarding how this action should be canceled. If not provided, then it is
    treated as though provided with `<CancelationMethodTerminate>`.
 
 The host uses the return code of the **command** run to determine success or failure of the Action. A zero exit code
 indicates success, and any non-zero exit code indicates failure. A timeout also indicates failure.
 
+When the WRAP_ACTIONS extension is in use, wrap scripts should signal infrastructure failure (Docker
```
mwiebe: Do we already have this kind of signal defined elsewhere? This doesn't feel specific to wrap actions, I think we'd either exclude this, or define this as a general thing that all actions can do.

wiki/How-Jobs-Are-Run.md
```diff
@@ -111,6 +121,12 @@ messages to convey information about the **Action** to the render management sys
   **Action** for entering an **Environment**. This unsets the given environment variable for all subsequent **Action**s
   in the **Session** until the **Environment** that emitted it is exited.
 
+When the WRAP_ACTIONS extension is in use, schedulers must scan the stdout of the wrap script (`onWrapEnter`,
+`onWrapTaskRun`, or `onWrapExit`) for these macros, not the stdout of the wrapped process. Wrap scripts must
```
mwiebe: Do we need to specify this? The scheduler has no visibility to the inside process, there might be multiple or no inside processes at all. Because of that I don't think there are alternatives to what's being described.

leongdl: I think the word scheduler is confused here, how about I just say "forward the macros" so the wrapped process's messages are still emitted by the wrapper so we do not lose the content.?

wiki/2023-09-Template-Schemas.md
leongdl: Sounds good, that makes sense.

wiki/2023-09-Template-Schemas.md
leongdl: Interesting - I was thinking about timeouts and how to handle docker general failures such as cannot start shared device, and found this 125 behavior. We can definitely cut it for now, and if we have a specific need to distinguish a wrapper failure vs task failure we can introduce this. 

(Seemed like an interesting add, but can also be a single sentence for the future to consider.)

rfcs/0008-environment-wrap-actions.md
leongdl: Sounds good, I'll refactor to just `WrappedAction.*`

rfcs/0008-environment-wrap-actions.md
leongdl: We can add f-format strings later since it is a programming improvements. 
After going through more examples, I actually think let bindings is a great way, since we can print them out, not recompute the variable etc. This can be a great programmer guide and suggestion.

Will update with `repr_sh`, good suggestion. I guess it is useful mainly on CMF, since SMF we use a simpler directory.

rfcs/0008-environment-wrap-actions.md
leongdl: True - given the comment above, I'll align the variables to `WrappedAction` then this example can be shorter. Its still flexible to have specific scripts per hook, so usability wise it is good.

rfcs/0008-environment-wrap-actions.md
leongdl: Yeah, that's definitely an option. As long as it meets our tenet of 1 wrapped action at most in the whole structure. 

I'll add a 1 sentence to keep it short, and a conformance test as example.

rfcs/0008-environment-wrap-actions.md
leongdl: Good point :) Now a steering rule in my own skills!

rfcs/0008-environment-wrap-actions.md
leongdl: Yes - I have applied this change to remove "on the host" generally throughout this document.

rfcs/0008-environment-wrap-actions.md
leongdl: Sounds good! Will remove the `>`

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+
+The following variables are read-only context supplied by the runtime to wrap
+actions. The `Action` segment of each name indicates that the variable corresponds
+to a field of the wrapped `<Action>` object. This allows new `<Action>` fields
+added in future specifications to be surfaced here without bespoke plumbing (see
+[Forwarding future `<Action>` fields](#forwarding-future-action-fields)).
+
+**Available in `onWrapEnter` and `onWrapExit`:**
+
+| Variable                          | Type           | Description |
```
leongdl, line 298: We have decided to unify the variables as per above.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+and `Task.Action.Environment` contain only the variables defined by
+`openjd_env` macros in the current session. They do NOT contain:
+
+- Host-inherited variables such as `HOME`, `USER`, `LANG`, `TZ`, or `PATH`
+  values that exist on the host but were never emitted via `openjd_env`.
+- OpenJD-injected variables such as `OPENJD_*` paths and session metadata.
+- Variables set by the wrapping environment's own `onEnter` through any
+  channel other than `openjd_env` stdout macros.
+
+Wrap scripts that need any of these variables in the wrapped execution
```
leongdl: Try to shorten this paragraph and section, it is defining only forwarding semantics.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+context MUST forward them explicitly. For Docker, this typically means
```
leongdl: Docker is "example", no need to be so wordy. The RFC should be agnostic of docker.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+adding `-e HOME` (forwards from host), `--env-file ...`, or per-variable
+`-e KEY=value` flags to the `docker container exec` invocation.
+
+**Embedded file paths and host filesystem references.** The wrapped action's
+`Command` and `Args` may reference host filesystem paths, most commonly
+embedded files (`{{Env.File.X}}` resolves to a host path), session working
+directory entries, or job-attachment paths. The wrap environment MUST ensure
+that the wrapped process can resolve every such path, typically by
```
leongdl: We can shorten this - docker and containers is an example to use bind mount and maintain path equivalence.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
++ If exactly one Environment in the session stack defines the wrap hooks (onWrapEnter,
++ onWrapTaskRun, onWrapExit), then the lifecycle actions of inner Environments and
++ tasks are intercepted:
++
++ - An inner Environment's onEnter is replaced by the wrapping Environment's onWrapEnter.
++ - A task's onRun is replaced by the wrapping Environment's onWrapTaskRun.
++ - An inner Environment's onExit is replaced by the wrapping Environment's onWrapExit.
++
++ The wrapping Environment's own onEnter and onExit are never wrapped; they always run
++ on the host. If more than one Environment in the session stack defines any wrap hook,
```
leongdl: nit, replace "on the host" as per above.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
++ the session is invalid and the scheduler must reject it before entering any
++ Environment. If an environment defines any wrap hook, it must define all three.
+```
+
+### Stdout forwarding and macro propagation
+
+The wrap script is the scheduler's direct child; the wrapped process is a
```
leongdl: Double check the use of "scheduler" here - does the session runner make more sense?

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+grand-child (see [Terminology](#terminology)). The two halves of this section
+specify the channel between them and the contract that flows through it.
+
+#### Stdout forwarding (channel)
+
+Wrap scripts MUST forward the wrapped process's stdout and stderr to their own
+stdout and stderr verbatim, without buffering, filtering, or transformation.
+This is the default behavior of `docker container exec`, `apptainer exec`, and
```
leongdl: for example, keep it shorter

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+`ssh` without options that suppress output, so the reference wrap scripts in
+[Appendix A](#appendix-a-reference-wrapper-scripts) inherit it automatically.
+Authors of custom wrap scripts (remote execution, privilege shifts,
+programmatic wrappers) MUST preserve this forwarding.
+
+#### Macro propagation (contract)
+
+Schedulers MUST scan the wrap script's stdout for OpenJD stdout macros: the
```
leongdl: NOt the scheduler - this should be transparently propagated via logging as per previous section. The OJD session runtime detects the `openjd_` macros and applies them normally. 

Given the wrapper, logs are pass through, and wrapped task's macro usage are unchanged and applied to the job as per the spec.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+`openjd_env`, `openjd_fail`, `openjd_progress`, and `openjd_status` macros
+defined in [How Jobs Are Run](https://github.com/OpenJobDescription/openjd-specifications/wiki/How-Jobs-Are-Run),
+along with any future macros added by later specifications. Schedulers MUST
+NOT scan the grand-child's stdout directly. This preserves the existing
```
leongdl: Rephrase or remove this - the word scheduler and scan here does is not correct.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+`Env.Wrapped.Action.Environment` every `openjd_env`-defined variable that
+was emitted on the stdout of any earlier action in the same session,
+including:
+
+- Variables emitted by host-context actions (the wrapping environment's own
+  `onEnter`, or any non-wrapped environment in the session stack).
+- Variables emitted by wrapped actions (an inner environment's `onEnter`
+  running via `onWrapEnter`, or earlier wrapped task `onRun` invocations).
+
+This is the chain that makes wrap composition useful: an inner Conda
```
leongdl: example: an inner Conda

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+#### Wrap-script-emitted macros
+
+A wrap script MAY emit stdout macros directly on its own stdout. Examples
+include `openjd_env: DOCKER_CONTAINER_ID=...` to publish a container ID
+established by `onEnter`, or `openjd_fail: <reason>` to declare an
+infrastructure failure with a human-readable message. The
+[Basic Example](#basic-example) uses this pattern to publish the container
+ID for use in subsequent wrap hooks.
+
+#### Alternatives considered
```
leongdl: Just remove the alternative considered for openjd_env propagation. Move it to out of scope above and propose current semantics of the spec apply and pass through the wrap. (Or consider the use of transitive?)

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+- Programmatic wrappers: return the wrapped process's exit code directly
+  (e.g., `sys.exit(proc.returncode)` in Python).
+
+#### Workload failure vs. infrastructure failure
+
+A wrap script can fail for two distinct reasons:
+
+1. **Workload failure**: the wrapped action itself exited non-zero. The
+   job logic is at fault.
+2. **Infrastructure failure**: the wrap script could not deliver the
```
leongdl: As per discussion, we combine infrastructure failure and workfload failure. Return non-zero and keep it simple. The exit code 125 can be moved to a future consideration if separating the failure mode is important.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+**Collision risk.** No operating system reserves exit status 125, so a
+workload that exits 125 is indistinguishable from an infrastructure failure.
+Wrap-script authors anticipating this collision SHOULD remap the workload's
+125 to a different non-zero code before propagating.
+
+The specification defines only the single infrastructure-failure category.
+Finer-grained classification (network vs. daemon vs. image-pull) is deferred
+to [Future Work](#future-work).
+
+#### Scheduler policy
```
leongdl: We can remove this section #### scheduler policy since it is superfluous.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+
+**Wrapping environment cleanup.** The wrapping environment's own `onEnter`
+and `onExit` are never wrapped; they run on the host. The existing OpenJD
+guarantee applies unchanged: the wrapping environment's `onExit` MUST run
+whenever the session entered (or attempted to enter) it, regardless of how
+any wrapped action terminated. This is the host-side teardown phase that
+stops the container, closes the remote session, releases the bind-mount, or
+otherwise releases the host resources allocated by `onEnter`.
+
+**Summary of run-on-failure guarantees:**
```
leongdl: Lead with this table and keep each description above shorter. Can make it easier to consume as a reader.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+
+**Signal propagation.** When the wrap script receives a termination signal
+(SIGTERM on POSIX, the platform-equivalent on Windows), it MUST cause the
+wrapped process to receive a termination signal within the wrap hook's
+`<Cancelation>` grace period. Non-propagation MAY produce orphaned processes,
+leaked container state, or stranded remote sessions.
+
+Any of the following idioms satisfies the requirement:
+
+- Shell wrap scripts SHOULD `trap` SIGTERM and forward it, or `exec` the
```
leongdl: Can we simplify this section - signal and error code propagation is the key and focus The inner task or enter exit fails, it is propagated by the wrap script. Keep 1 example with a sig term, and 1 with docker. Make this shorter.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,1497 @@
...
+Sentinel `0` was chosen because the OpenJD `<posinteger>` schema cannot
+represent absent values, and `0` is otherwise meaningless as a positive
+timeout duration.
+
+**Relation to cancelation.** The scheduler enforces the hook timeout by
+applying the wrap hook's `<Cancelation>` policy. Signal propagation from
+the wrap script to the wrapped process is normative and specified in
+[Cancelation behavior](#cancelation-behavior).
+
+## Design Choice Rationale
```
leongdl, line 484: This can be an appendix. Its good for discussion, maybe can be deleted in the final version.

rfcs/0008-environment-wrap-actions.md
leongdl: Got it - will rename.

rfcs/0008-environment-wrap-actions.md
leongdl: That is true - the spec can only give the best practice. Our examples is the demonstrator, but the user is the one is responsible to re-use our examples. Will do!

rfcs/0008-environment-wrap-actions.md
leongdl: Makes sense, too much details :)

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-ignores-wrapped-action-vars.test.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - EXPR
+  environment:
+    name: WrapEnv
+    script:
+      actions:
+        onWrapEnter:
+          command: sh
+          args:
+          - "-c"
+          - "{{ repr_sh(WrappedAction.Command) }} {{ repr_sh(WrappedAction.Args) }}"
```
leongdl, line 28: Curious if this should "fail" - since the enter is not wrapped.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,820 @@
...
+
+4. **Privilege shifts.** Run inner actions as a different user or with reduced
+   capabilities by wrapping the command with `sudo -u`, `unshare`, or similar.
+
+### Backward compatibility
+
+This RFC is additive and gated by the `WRAP_ACTIONS` extension name declared under
+RFC 0002. Specifically:
+
+- Schedulers that do not implement `WRAP_ACTIONS` MUST reject templates that list
```
leongdl, line 186: Curious about the use of the word "schedulers", is there a specific verb we call openJD runtimes?

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,820 @@
...
+`--timeout 0`, which Docker interprets as "kill immediately").
+The sentinel `0` was chosen because the OpenJD `<posinteger>` schema cannot
+represent absent values.
+
+Templates that omit `timeout` on a wrap hook inherit OpenJD's default for
+that hook position. The scheduler enforces the hook timeout by applying the
+wrap hook's `<Cancelation>` policy (see
+[Cancelation behavior](#cancelation-behavior)).
+
+## Design Choice Rationale
```
leongdl, line 484: Curious if we would like to keep this section - for brevity.

conformance-tests/2023-09/WRAP_ACTIONS/env_templates/4.2--empty-actions-no-wrap-no-enter-no-exit.invalid.yaml
```diff
@@ -0,0 +1,9 @@
+# An environment whose `actions` block is empty is invalid: at least one of
+# onEnter, onWrapEnter, onWrapTaskRun, onWrapExit, or onExit must be defined.
+specificationVersion: environment-2023-09
+extensions:
+- WRAP_ACTIONS
```
mwiebe, line 5: I think this should already fail because EXPR is missing, so it doesn't test what the comment says.

conformance-tests/2023-09/WRAP_ACTIONS/env_templates/4.2--minimal-wrap-env-template.yaml
```diff
@@ -0,0 +1,23 @@
...
+environment:
+  name: MinimalWrapEnv
+  script:
+    actions:
+      onEnter:
+        command: echo
+        args: ["entered"]
+      onWrapEnter:
+        command: echo
+        args: ["wrap-enter", "{{Env.Wrapped.Name}}"]
```
mwiebe: Going with `WrappedEnv.Name` (and likely `WrappedStep.Name` inside `onWrapTaskRun` is probably more consistent with the `WrappedAction` convension.

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-empty-environment-list.test.yaml
```diff
@@ -0,0 +1,54 @@
...
+            done
+            echo ENV_DUMP_COMPLETE
+        onWrapExit:
+          command: sh
+          args:
+          - "-c"
+          - "{{ repr_sh(WrappedAction.Command) }} {{ repr_sh(WrappedAction.Args) }}"
+
+runOn:
+- posix
```
mwiebe: If this isn't posix-specific, better to run on all OS.

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-empty-environment-list.test.yaml
```diff
@@ -0,0 +1,54 @@
...
+- specificationVersion: environment-2023-09
+  extensions:
+  - WRAP_ACTIONS
+  - EXPR
+  environment:
+    name: WrapEnv
+    script:
+      actions:
+        onWrapEnter:
+          command: sh
```
mwiebe: Suggestion change this to `bash` as that's what we generally have on Windows as well.

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-enter-and-exit-three-hooks.test.yaml
```diff
@@ -0,0 +1,62 @@
...
+          args: ["INNER_EXIT_NORMAL"]
+  steps:
+  - name: Step1
+    script:
+      actions:
+        onRun:
+          command: echo
+          args: ["ORIGINAL_TASK"]
+
+environments:
```
mwiebe, line 26: Something to double check: Make sure we're testing with wrapping environment inside the job (both as job environments and step environments) as well. I'm not very far in so good chance they're later!

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-enter-and-exit-three-hooks.test.yaml
```diff
@@ -0,0 +1,62 @@
...
+  - EXPR
+  environment:
+    name: WrapEnv
+    script:
+      actions:
+        onWrapEnter:
+          command: sh
+          args:
+          - "-c"
+          - "echo WRAP_ENTER_NAME={{Env.Wrapped.Name}}"
```
mwiebe: Could also echo the WrappedAction.Command here, and then the Step Name in the wrap task run.

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-glob-characters-literal.test.yaml
```diff
@@ -0,0 +1,50 @@
...
+          args:
+          - "-c"
+          - "{{ repr_sh(WrappedAction.Command) }} {{ repr_sh(WrappedAction.Args) }}"
+
+runOn:
+- posix
+
+expected:
+  output:
+  - "ARG=*.txt"
```
mwiebe, line 48: This doesn't feel right as the output for that command, shouldn't it be more like

```
ARG='*.txt' "?.log" "[abc*"
```

because I only see one ARG= echo. Probably confusing `sh` details.

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-ignores-wrapped-action-vars.test.yaml
mwiebe: It looks valid to me because that's syntax check, not a runtime check that the template it's applied against actually has an onEnter.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,795 @@
...
+
+### Pydantic WrapValidator
+
+The naming and concept are inspired by
+[Pydantic's WrapValidator](https://docs.pydantic.dev/latest/concepts/validators/#wrap-validators),
+which wraps a validation step with before/after logic. The wrap hooks wrap the
+lifecycle actions of inner environments and tasks with before/after logic
+provided by the outer environment.
+
+### Workflow DSLs with container abstraction
```
mwiebe: Maybe we skip this part, it's not actually about wrapping.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,795 @@
...
+- **CWL** declares `DockerRequirement` as a hint; runners translate at execution.
+- **Snakemake** declares `container:` per rule and converts at runtime.
+- **WDL** declares `docker:` per task; engines may substitute.
+
+OpenJD's approach uses composable environment templates. The distinction from the
+workflow-DSL prior art is that OpenJD separates the outer execution-context layer
+(the wrap environment) from the job-logic layer (the job template), rather than
+annotating individual processes with a runtime identifier.
+
+### Conda and Rez run commands
```
mwiebe: Could include Docker here too, its command is similar.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,795 @@
...
+[Conda](https://docs.conda.io/projects/conda/en/latest/commands/run.html)
+(`conda run -n <env> <cmd>`) and
+[Rez](https://rez.readthedocs.io/en/stable/commands/rez-env.html)
+(`rez-env <pkgs> -- <cmd>`) both expose a "run a command inside this
+environment" pattern. A wrap environment that exec's `conda run` or `rez-env`
+in its wrap hooks reuses that idea at the session level.
+
+## Rejected Ideas
+
+### Extending `<Action>` with a `wrapper` field
```
mwiebe: Consider dropping this one.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,795 @@
...
+want per-phase schema explicitness; the two forms would be mutually exclusive
+per environment. Reasons for rejecting it from the initial RFC are in
+[Design Choice Rationale](#three-separate-hooks-rather-than-a-single-unified-hook).
+
+### Nested wrap composition
+
+Stacking multiple wrap environments (for example, a profiling wrapper outside a
+container wrapper) is valuable but adds implementation complexity. Deferred.
+
+### Wrap-aware cancelation
```
mwiebe: If this isn't too bad to implement, maybe we want to add this in?

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,795 @@
...
+failures.
+
+### Session health monitoring
+
+Periodic health-checking of wrapped processes (detecting stalled containers,
+unresponsive remote hosts) is a useful adjacent feature, but is cleanly separable
+from the wrap mechanic. A separate RFC for environment health monitoring is the
+appropriate venue.
+
+### Cross-OS wrapping
```
mwiebe: Consider dropping this one.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,795 @@
...
+  B.onEnter                          (via A.onWrapEnter)
+    Task 1: onRun                    (via A.onWrapTaskRun)
+    Task 2: onRun                    (via A.onWrapTaskRun)
+  B.onExit                           (via A.onWrapExit)
+A.onExit                             (wrapping env's own onExit is never wrapped)
+```
+
+The wrapping environment is not required to be the outermost. If `B` were the
+wrapping environment instead, `A` would run normally and `B` would wrap its
+own inner actions and tasks.
```
leongdl, line 128: Added to the Basic Example section, just below the execution-order diagram:

> The wrapping environment is not required to be the outermost. If `B` were the wrapping environment instead, `A` would run normally and `B` would wrap its own inner actions and tasks.

Thanks for flagging — useful context for readers.

leongdl (commented):
Posting replies and acknowledgements to a batch of comments. Latest revision on the branch addresses these.

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-glob-characters-literal.test.yaml
```diff
@@ -0,0 +1,50 @@
...
+        onWrapEnter:
+          command: sh
+          args:
+          - "-c"
+          - "{{ repr_sh(WrappedAction.Command) }} {{ repr_sh(WrappedAction.Args) }}"
+        onWrapTaskRun:
+          command: sh
+          args:
+          - "-c"
+          - "printf 'ARG=%s\\n' {{ repr_sh(WrappedAction.Args) }}"
```
leongdl, line 36: I think this output is actually correct — `printf` cycles its format string for any extra arguments beyond what the format consumes. So `printf 'ARG=%s\n' a b c` produces three lines:

```
ARG=a
ARG=b
ARG=c
```

Not one line with all three values. The expected three `ARG=...` lines (one per arg) match what the wrap script will actually emit. Quoted from `printf(1)`:

> The format string is reused as often as necessary to satisfy the arguments.

Let me know if I'm reading the test wrong, otherwise leaving as-is.

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-ignores-wrapped-action-vars.test.yaml
```diff
@@ -0,0 +1,45 @@
...
+environments:
+- specificationVersion: environment-2023-09
+  extensions:
+  - WRAP_ACTIONS
+  - EXPR
+  environment:
+    name: WrapEnv
+    script:
+      actions:
+        onWrapEnter:
```
leongdl: Agreed — the test is intentionally a positive (`.test.yaml`, not `.invalid.test.yaml`) checking that the wrap action runs successfully when it ignores the injected variables. No change needed; thanks for confirming.

leongdl (commented):
Replies to the conformance-test thread on wrap-glob-characters-literal.test.yaml and wrap-ignores-wrapped-action-vars.test.yaml.

conformance-tests/2023-09/WRAP_ACTIONS/env_templates/4.2--empty-actions-no-wrap-no-enter-no-exit.invalid.yaml
leongdl: Fixed - made sure all templates hae both EXPR and WRAP_ACTIONS

wiki/2023-09-Template-Schemas.md
leongdl: Fixed in small patch updates. Rebasing to clean up the commit.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,759 @@
...
+* Accepted On: (pending)
+* Depends On:
+  * RFC 0002, Model Extensions (https://github.com/OpenJobDescription/openjd-specifications/issues/57)
+  * RFC 0005, Expression Language (https://github.com/OpenJobDescription/openjd-specifications/pull/93)
+  * RFC 0006, Expression Function Library (https://github.com/OpenJobDescription/openjd-specifications/pull/104)
+
+## Summary
+
+This RFC proposes extending `<Environment>` with three new session actions
+(`onWrapEnter`, `onWrapTaskRun`, and `onWrapExit`) that let an environment
```
mwiebe: There's a naming inconsistency that we should iron out between the three wrapping actions.

* onWrapTaskRun - wraps the Task `onRun` action
* onWrapEnter - wraps the Environment `onEnter` action
* onWrapExit - wraps the Environment `onExit` action

The first one includes "Task", representing the object that contains the action, while the second two don't include "Environment" or "Env". I think we should rename the latter two into `onWrapEnvEnter` and `onWrapEnvExit`.

This matters because we've talked about a kind of extension where there might be more object types that could have onEnter/onExit. If we add `jobServices` and `stepServices` analogous to the environments, they could then get their own `onWrapServiceEnter`/`onWrapServiceExit`.

leongdl: Great idea, why don't we settle on 

onWrapTaskRun - wraps the Task onRun action
onWrapEnvEnter - wraps the Environment onEnter action
onWrapEnvExit - wraps the Environment onExit action

rfcs/0008-environment-wrap-actions.md
mwiebe: "Schedulers" seems fine to me, and "runtimes" could work too.

leongdl:
As per offline chats - we are moving this RFC to final comments.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,759 @@
...
+### Backward compatibility
+
+This RFC is additive and gated by the `WRAP_ACTIONS` extension name declared under
+RFC 0002. Specifically:
+
+- Schedulers that do not implement `WRAP_ACTIONS` MUST reject templates that list
+  it in `extensions:`, per RFC 0002's extension-handling rules.
+- Schedulers that do implement `WRAP_ACTIONS` MUST ignore the three wrap hooks on
+  templates that do not list `WRAP_ACTIONS` in `extensions:`, so that existing
+  templates continue to behave exactly as before.
```
mwiebe: Is this about the case where an external environment uses `WRAP_ACTIONS` but the job template doesn't? In that case I think the scheduler must not ignore the wrap hooks. This bullet point sounds wrong.

rfcs/0008-environment-wrap-actions.md
leongdl: Ok. I will keep them as is.

rfcs/0008-environment-wrap-actions.md
leongdl: Yes exactly - Let me reframe this to be more clear.

Environment -> has wrap action
Jobs -> no extensions  (Classic or existing jobs)

When running the job it should RUN the hooks.

rfcs/0008-environment-wrap-actions.md
leongdl: Sounds good, I'll delete it.

rfcs/0008-environment-wrap-actions.md
leongdl: Sounds good, was an extra idea, deleted.

rfcs/0008-environment-wrap-actions.md
leongdl: Sure, I can move it back and then update the code. Should be easy

leongdl: Will ship a separate PR to openjd-rs and openjd-model-for-python to add this. Since the current prototype is already merged.

leongdl:
@mwiebe re your question on the `list[string]` env var handling ("Can you check if `e.split('=', 1)` works as expected to be able to separately access the key and value?"):

Yes — confirmed. `e.split('=', 1)` works and returns `["KEY", "value"]`, splitting on the first `=` only so values containing `=` are preserved.

Verified against both sources:
- **RFC 0006** defines `split(s: string, sep: string, maxsplit: int) -> list[string]`, and documents that the method form `"a,b,c".split(",")` is equivalent to the function form `split("a,b,c", ",")`.
- **openjd-rs** reference implementation (`openjd-expr` crate): `split_fn` uses `s.splitn(maxsplit + 1, sep)`, and method calls desugar to the function call, so `e.split('=', 1)` evaluates identically to `split(e, '=', 1)`.

Indexing/slicing are also supported, so `e.split('=', 1)[0]` and `e.split('=', 1)[1]` give the key and value respectively.

I've added a note to the `list[string]` rationale in the RFC pointing this out.

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-exit-status-windows.test.yaml
```diff
@@ -0,0 +1,44 @@
...
+          - "{{ repr_cmd(WrappedAction.Command) }} {{ repr_cmd(WrappedAction.Args) }}"
+        onWrapEnvExit:
+          command: cmd
+          args:
+          - "/c"
+          - "{{ repr_cmd(WrappedAction.Command) }} {{ repr_cmd(WrappedAction.Args) }}"
+
+expected:
+  taskFailure:
+    exitCode: 42
```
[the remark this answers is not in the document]
leongdl: I will add support for this separately

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-exit-status-python.test.yaml
```diff
@@ -0,0 +1,42 @@
...
+          - "-c"
+          - "import subprocess,sys; sys.exit(subprocess.run([{{repr_py(WrappedAction.Command)}}]+{{repr_py(WrappedAction.Args)}}).returncode)"
+        onWrapEnvExit:
+          command: python
+          args:
+          - "-c"
+          - "import subprocess,sys; sys.exit(subprocess.run([{{repr_py(WrappedAction.Command)}}]+{{repr_py(WrappedAction.Args)}}).returncode)"
+
+expected:
+  taskFailure:
```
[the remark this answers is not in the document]
leongdl: I will add support for this separately.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -0,0 +1,744 @@
...
+
+**Available in `onWrapEnvEnter`, `onWrapTaskRun`, and `onWrapEnvExit`:**
+
+| Variable                          | Type           | Description |
+|-----------------------------------|----------------|-------------|
+| `WrappedAction.Command`           | `string`       | The `command` from the wrapped action. |
+| `WrappedAction.Args`              | `list[string]` | The `args` from the wrapped action. |
+| `WrappedAction.Environment`       | `list[string]` | Environment variables defined by `openjd_env` earlier in the session, as `["KEY=value", ...]`. See [Host environment variables and embedded file paths](#host-environment-variables-and-embedded-file-paths). |
+| `WrappedAction.Timeout`           | `int`          | The timeout in seconds specified on the wrapped action, or `0` if none. See [Timeout behavior](#timeout-behavior). |
+| `WrappedAction.Cancelation.Mode`  | `string`       | The cancelation method of the wrapped action (`TERMINATE` or `NOTIFY_THEN_TERMINATE`), or empty if the wrapped action defines no `<Cancelation>`. See [Cancelation behavior](#cancelation-behavior). |
```
mwiebe, line 294: This is adding an extra `.Mode` here that doesn't exist in the `<Action>` definition itself: https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#5-action

mwiebe: Sorry, should have followed to https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#53-cancelationmethod. As a next step we should look at how a template could forward the full Cancelation object instead of just its Mode.

wiki/2023-09-Template-Schemas.md
```diff
@@ -1502,6 +1503,16 @@ The format string scopes available to format strings within an Environment are:
 3. `Env.*` — Scope of the environment entity itself. Values such as the embedded files defined within the Environment
    entity.
 4. Names bound by `let` in the enclosing `<EnvironmentScript>`. Available with the `EXPR` extension.
+5. `WrappedAction.*` — Available within `onWrapEnvEnter`, `onWrapTaskRun`, and `onWrapEnvExit` only. Carries
+   the wrapped `<Action>`'s fields. Available with the `WRAP_ACTIONS` extension. See [RFC 0008](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0008-environment-wrap-actions.md).
```
mwiebe, line 1507: This spec doc is the authoritative reference, so we should copy that here instead of referencing the RFC.

mwiebe (approved):
I see some references from the schema to the RFC for things that should live in the schema. Let's merge and then do a follow-up where we fix that.

Issue #134, opened 2026-04-22
RFC: Range-based capability matching for amount requirements


oehmends:
* **Pull Request**: https://github.com/OpenJobDescription/openjd-specifications/pull/135
* **Discussion Thread(s)**:
    * Related AWS Support case: Fleet compatibility check matches step requirements against fleet minimum vCPU instead of maximum, forcing users to maintain separate fleets per resource tier.

## Description

The OpenJD specification defines the data model for step amount requirements (`<AmountRequirement>` with `min`/`max`) and worker capabilities, but does not define how scheduling systems should match requirements against hosts or fleets that advertise a **range** of capabilities (e.g., vCPU min–max).

This gap has led to implementations that match only against the lower bound of a capability range. For example, a fleet configured with vCPU range 64–128 is rejected for a step requiring 96 vCPUs, because the check compares 96 against the fleet minimum (64) rather than the fleet maximum (128).

This forces users into one of two bad workarounds:
1. **Fleet proliferation** — Create separate fleets for each vCPU/memory tier (64, 96, 128), each requiring its own service quotas, monitoring, and configuration.
2. **Over-provisioning** — Set the fleet minimum to the highest value any job might need, wasting resources on smaller jobs.

Both defeat the purpose of flexible fleet sizing and increase operational overhead.

This RFC proposes normative language defining that a fleet with capability range [C_min, C_max] should be considered compatible with a step requirement if there exists any value in the fleet's range that satisfies the requirement. In practice: the fleet's **maximum** determines whether it can meet the step's minimum requirement.

## Roles

| Role | User
| ---- | ----
| Proposed By | @oehmends
| Author(s)   | @oehmends

## Workflow

- [x] Tracking issue created (label: `rfc/proposed`)
- [ ] RFC pull request submitted and ready for discussion (label: `rfc/exploring`)
- [ ] Last call for comments (labels: `rfc/exploring` and `rfc/final-comments`)
- [ ] Accepted and merged RFC pull request (label: `rfc/accepted-future`)
- [ ] Green-light for inclusion in a draft specification, and the author is creating and iterating on pull requests (label: `rfc/accepted-draft`)
- [ ] Pull requests are merged in to a draft specification (label: `rfc/accepted-staged`)
- [ ] Officially published in a non-draft revision of the specification (label: `rfc/released`)

Please close this tracking issue when the proposal enters the `Released` stage of the process.

## Open Points

[...]

PR #135, opened 2026-04-22
feat: RFC for range-based capability matching for amount requirements

oehmends:
Tracking Issue: #134

## Summary

This RFC proposes normative language defining how scheduling systems should match step amount requirements against hosts or fleets that advertise a **range** of capabilities.

Currently, the spec defines the data model for requirements and capabilities but leaves matching semantics undefined. This has led to implementations that match against the lower bound of a fleet's capability range — a fleet with vCPU range 64–128 is rejected for a step requiring 96 vCPUs because 96 > 64 (the fleet minimum).

This forces users to either maintain separate fleets per resource tier (each with its own quotas) or over-provision by setting fleet minimums to the highest value any job might need. Both defeat the purpose of flexible fleet sizing.

## Proposed change

A fleet with capability range [C_min, C_max] is compatible with a step requirement if there exists any value in the range that satisfies the requirement:
- Fleet max (C_max) >= step min requirement (can the fleet provide enough?)
- Fleet min (C_min) <= step max requirement (can the fleet stay small enough?)

## What this enables

- A single fleet with vCPU range 64–128 serves jobs requiring 64, 96, or 128 vCPUs
- No fleet proliferation, no extra quotas, no over-provisioning
- Consistent, predictable behavior across scheduling system implementations

### DIFF
--- a/rfcs/0008-range-based-capability-matching.md
+++ b/rfcs/0008-range-based-capability-matching.md
@@ -1,0 +1,240 @@
+
+* Feature Name: range_based_capability_matching
+* Author(s): Johannes Oehmen `[oehmends](https://github.com/oehmends)`
+* RFC Tracking Issue: https://github.com/OpenJobDescription/openjd-specifications/issues/134
+* Start Date: 2026-04-22
+* Specification Version: 2023-09
+* Accepted On: (pending)
+
+## Summary
+
+This RFC proposes adding normative language to the Open Job Description specification that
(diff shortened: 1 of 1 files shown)


oehmends:
Closing this RFC — per discussion in #134, fleet-level matching semantics are intentionally outside the scope of the OpenJD spec and belong as a Deadline Cloud feature request. Thanks @mwiebe for the guidance.

PR #137, opened 2026-04-24
feat: Add OPENJD_SESSION_WORKING_DIR environment variable to spec

seant-aws:
## PR for other purpose

Add documentation and conformance tests for the OPENJD_SESSION_WORKING_DIR environment variable, which exposes the Session working directory to nested subprocesses.

Documentation changes:
- New 'Session Environment Variables' section in How-Jobs-Are-Run.md with design rationale explaining why env vars are generally avoided (tooling parseability) and why this one is acceptable (same value as template var, CWD already points there)
- Updated Session.WorkingDirectory description in How-Jobs-Are-Constructed.md to reference the corresponding environment variable

Conformance tests:
- Verify OPENJD_SESSION_WORKING_DIR is set and matches Session.WorkingDirectory
- Verify the env var propagates to child processes

---


wiki/How-Jobs-Are-Run.md
```diff
@@ -45,6 +45,17 @@ The following diagram provides an overview of the steps taken to run a Session.
 
 ![Actions Run in a Session](./images/2023-09_how_jobs_are_run.png)
 
+## Session Environment Variables
```
mwiebe, line 48: This section doesn't include the rationale details that the PR description claims it does.

mwiebe (changes requested):

wiki/How-Jobs-Are-Run.md
seant-aws: My mistake, added another section below. Thanks !

mwiebe (approved):

PR #148, opened 2026-07-15
feat: Polish up RFC 0008 and address follow-up review comments

leongdl:
## Summary

Follow-up to #130, addressing the review comments that were deferred at merge time. RFC 0008 (Environment Wrap Actions) was approved with: *"I see some references from the schema to the RFC for things that should live in the schema. Let's merge and then do a follow-up where we fix that."*

This PR resolves that condition plus the remaining open review threads.

### 1. Schema doc delegates normative content to the RFC (@mwiebe, approval condition + inline on `2023-09-Template-Schemas.md` L1507)

> "This spec doc is the authoritative reference, so we should copy that here instead of referencing the RFC."

**Addressed by** copying the normative content into the spec docs and keeping RFC links for attribution only:

- `2023-09-Template-Schemas.md`: new section **4.3.1 Wrap hook template variables** with the full `WrappedAction.*` / `WrappedEnv.Name` / `WrappedStep.Name` tables, the host-env/path forwarding rule, the `openjd_env` accumulation rule, and the future-fields PascalCase namespace rule. The `<Action>` and `<SimpleAction>` timeout footnotes now carry the wrap-script timeout propagation and zero-sentinel rules instead of linking out.
- `How-Jobs-Are-Run.md`: copied in wrap failure semantics (a wrap hook is treated as the action it replaces; exit-status propagation), the symmetric cleanup guarantee, cancelation signal propagation, and the stdout forwarding / macro-scanning rules.

### 2. Forwarding the full Cancelation object (@mwiebe, inline on RFC L294)

> "As a next step we should look at how a template could forward the full Cancelation object instead of just its Mode."

**Proposal included in this PR:** add `WrappedAction.Cancelation.NotifyPeriodInSeconds` (`int?`) alongside the existing `WrappedAction.Cancelation.Mode`:

- When the wrapped action's mode is `NOTIFY_THEN_TERMINATE`, the runtime supplies the *effective* grace period, applying the schema defaults (120 for a task's `onRun`, 30 otherwise) when the field is omitted — so the wrap script always sees the value the runtime would have enforced in the unwrapped case.
- `null` when the mode is `TERMINATE` or no `<Cancelation>` is defined — a notify period does not apply in those cases, and typing the variable as `int?` avoids conflating "not applicable" with a zero-length notify period. *(Updated per review feedback — earlier revisions used a `0` sentinel matching the `WrappedAction.Timeout` convention.)* Because `null` renders as the empty string in a format string, the RFC now documents the guard needed for unconditional references: branch on `WrappedAction.Cancelation.Mode` first, or use EXPR null-coalescing, e.g. `docker container stop --timeout {{ WrappedAction.Cancelation.NotifyPeriodInSeconds or 30 }}`.

**Alternative considered:** forwarding the whole `<Cancelation>` as a structured object. Rejected for ergonomics — it adds another level of depth to the `WrappedAction.Cancelation` chain without carrying more information, and scalar variables compose directly with `repr_sh()`/`repr_py()` and EXPR arithmetic in wrap scripts. Future `<CancelationMethod>` fields surface as additional PascalCased scalars under `WrappedAction.Cancelation.*`, per the RFC's existing future-fields rule. These two variables cover every field of today's `<CancelationMethod>` schema.

### 3. Basic Example uses invalid format-string timeouts (automated review, unresolved)

> `timeout: "{{WrappedAction.Timeout}}"` conflicts with the schema: format-string timeouts require `FEATURE_BUNDLE_1` (not declared in the example), and `WrappedAction.Timeout` evaluates to `0` when the wrapped action has no timeout, which is out of range for `<posinteger>`.

**Addressed by** removing the `timeout:` lines from the example's wrap hooks (they now inherit the defaults for their hook positions) and adding a paragraph after the example explaining why, pointing at the Timeout behavior section for how the wrapped timeout is enforced from inside the wrap script instead.

A new invalid fixture guards the mistake: `4.2--wrap-fmtstring-timeout-no-feature-bundle.invalid.yaml` asserts a format-string `timeout` on a wrap hook is rejected without `FEATURE_BUNDLE_1`.

### 4. `wrap-partial-hooks-rejected` could fail for the wrong reason (automated review, unresolved)

> The env declares only `WRAP_ACTIONS` and omits `EXPR`, so a validator may reject it on the missing-`EXPR` rule before ever evaluating the all-or-nothing rule the test is meant to exercise.

**Addressed by** adding `EXPR` to the environment's extensions so the partial hook set is the only remaining defect.

### 5. README timeout-sentinel bullet mislists a fixture (automated review, unresolved)

> `wrap-timeout-injected-python` sets `timeout: 300` and asserts `TIMEOUT=300` — it does not exercise the `0`-when-unset case.

**Addressed by** moving it under the "Variable injection" bullet; the zero-sentinel bullet now references only `wrap-task-action-timeout-zero-when-unset`.

## Conformance coverage for the new variables

Nine new job fixtures pin the `WrappedAction.Cancelation.*` semantics (see the WRAP_ACTIONS conformance README for the full inventory):

- **Mode injection**: `TERMINATE`, `NOTIFY_THEN_TERMINATE`, and the empty string when the wrapped action defines no `<Cancelation>` (three fixtures — the empty case is distinct from an explicit `TERMINATE` so implementations cannot conflate "author declared TERMINATE" with "author declared nothing").
- **Notify-period injection**: explicit value forwarded verbatim; schema-default forwarding for all three wrapped-action positions (120 for a task's `onRun`; 30 for an env `onEnter` and an env `onExit`); and the `null` sentinel for both not-applicable cases (`TERMINATE` mode, and no `<Cancelation>` declared).

## Testing

- YAML fixtures validated for syntax; conformance CI on this PR exercises the new/updated tests against both the Python and Rust CLIs.
- The new `4.2--wrap-fmtstring-timeout-no-feature-bundle.invalid.yaml` fixture doubles as a check that both validators reject format-string wrap-hook timeouts without `FEATURE_BUNDLE_1`.
- **Rust (openjd-rs)**: the `Cancelation.*` seed-path and validator changes are implemented (leongdl/openjd-rs `1fd8ea5`); all 9 new fixtures pass, and the full local conformance run is 1138 passed / 0 failed.
- **Python (openjd-sessions-for-python)**: implements the PR #130 variables (`WrappedAction.Command/Args/Environment/Timeout`) but does not yet seed `WrappedAction.Cancelation.*`. The matching port (session seed path + validator symbol registration) is needed before conformance CI can go green on both CLIs.

---

### DIFF
--- a/rfcs/0008-environment-wrap-actions.md
+++ b/rfcs/0008-environment-wrap-actions.md
@@ -55,15 +55,12 @@ environment:
       onWrapEnvEnter:
         command: bash
         args: ["{{Env.File.Wrap}}"]
-        timeout: "{{WrappedAction.Timeout}}"
       onWrapTaskRun:
         command: bash
         args: ["{{Env.File.Wrap}}"]
-        timeout: "{{WrappedAction.Timeout}}"
       onWrapEnvExit:
         command: bash
         args: ["{{Env.File.Wrap}}"]
-        timeout: "{{WrappedAction.Timeout}}"
       onExit:
         command: bash
         args: ["{{Env.File.Exit}}"]
@@ -103,6 +100,15 @@ environment:
           set -euo pipefail
           docker container stop --timeout 30 "$DOCKER_CONTAINER_ID"
 ```
+
+The wrap hooks declare no `timeout`, so each inherits the default for its hook
+position (see [Timeout behavior](#timeout-behavior)). With the `FEATURE_BUNDLE_1`
+extension a hook may instead forward the wrapped action's timeout as its own
+via `timeout: "{{WrappedAction.Timeout}}"` — the variable is `int?`, so a
+declared timeout forwards verbatim and a `null` (no timeout on the wrapped
+action) drops the field, keeping the hook default. This example enforces the
+wrapped action's timeout from inside the wrap script instead (see
+[Timeout behavior](#timeout-behavior)).
 
 The job template that runs under this wrapping environment is unchanged from one
 that runs without wrapping. See the existing
@@ -290,8 +296,9 @@ helper scripts can be reused unchanged.
 | `WrappedAction.Command`           | `string`       | The `command` from the wrapped action. |
 | `WrappedAction.Args`              | `list[string]` | The `args` from the wrapped action. |
 | `WrappedAction.Environment`       | `list[string]` | Environment variables defined by `openjd_env` earlier in the session, as `["KEY=value", ...]`. See [Host environment variables and embedded file paths](#host-environment-variables-and-embedded-file-paths). |
-| `WrappedAction.Timeout`           | `int`          | The timeout in seconds specified on the wrapped action, or `0` if none. See [Timeout behavior](#timeout-behavior). |
-| `WrappedAction.Cancelation.Mode`  | `string`       | The cancelation method of the wrapped action (`TERMINATE` or `NOTIFY_THEN_TERMINATE`), or empty if the wrapped action defines no `<Cancelation>`. See [Cancelation behavior](#cancelation-behavior). |
+| `WrappedAction.Timeout`           | `int?`         | The timeout in seconds specified on the wrapped action. `null` when the wrapped action specifies no timeout — no bound is meaningful in that case. See [Timeout behavior](#timeout-behavior). |
+| `WrappedAction.Cancelation.Mode`  | `string?`      | The cancelation method of the wrapped action (`TERMINATE` or `NOTIFY_THEN_TERMINATE`). `null` when the wrapped action defines no `<Cancelation>` — the author declared nothing, so no method name is meaningful. See [Cancelation behavior](#cancelation-behavior). |
+| `WrappedAction.Cancelation.NotifyPeriodInSeconds` | `int?` | The effective `notifyPeriodInSeconds` of the wrapped action when its cancelation mode is `NOTIFY_THEN_TERMINATE`, with schema defaults applied when the wrapped action omits the field. `null` when the mode is `TERMINATE` or the wrapped action defines no `<Cancelation>` — a notify period does not apply to those cases, so no numeric value is meaningful. See [Cancelation behavior](#cancelation-behavior). |
 
 **Additionally available in `onWrapEnvEnter` and `onWrapEnvExit`:**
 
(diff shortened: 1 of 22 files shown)


rfcs/0008-environment-wrap-actions.md
```diff
@@ -292,6 +297,7 @@ helper scripts can be reused unchanged.
 | `WrappedAction.Environment`       | `list[string]` | Environment variables defined by `openjd_env` earlier in the session, as `["KEY=value", ...]`. See [Host environment variables and embedded file paths](#host-environment-variables-and-embedded-file-paths). |
 | `WrappedAction.Timeout`           | `int`          | The timeout in seconds specified on the wrapped action, or `0` if none. See [Timeout behavior](#timeout-behavior). |
 | `WrappedAction.Cancelation.Mode`  | `string`       | The cancelation method of the wrapped action (`TERMINATE` or `NOTIFY_THEN_TERMINATE`), or empty if the wrapped action defines no `<Cancelation>`. See [Cancelation behavior](#cancelation-behavior). |
+| `WrappedAction.Cancelation.NotifyPeriodInSeconds` | `int` | The effective `notifyPeriodInSeconds` of the wrapped action when its cancelation mode is `NOTIFY_THEN_TERMINATE`, with schema defaults applied when the wrapped action omits the field. `0` when the mode is `TERMINATE` or the wrapped action defines no `<Cancelation>`. See [Cancelation behavior](#cancelation-behavior). |
```
mwiebe: We could make this `int?` and be `null` when the mode is `TERMINATE`. I think this would play naturally with expanding it.

leongdl: Sounds good to me! Let me change it/

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-cancelation-mode-empty-when-no-cancelation.test.yaml
```diff
@@ -0,0 +1,50 @@
+# RFC 0008: WrappedAction.Cancelation.Mode is the empty string when the
+# wrapped action defines no <Cancelation>. Note this differs from the
+# runtime's default cancelation-method semantics: at runtime, an action with
+# no <Cancelation> is treated as `<CancelationMethodTerminate>` (Template
+# Schemas 5), but the template variable itself carries the "not declared"
+# sentinel — the empty string — so wrap scripts can distinguish "author
```
mwiebe: Using an empty string isn't following the semantics that EXPR defined for optional data, can you change the spec so that it's `null`, and this variable is of type `string?` instead of `string`? Similar to the notify period change.

leongdl: Thanks! Updated:
```
Changed WrappedAction.Cancelation.Mode from string with an empty-string sentinel to string? that is null when the wrapped action defines no <Cancelation>, following the EXPR semantics for optional data and mirroring the earlier int? change to Cancelation.NotifyPeriodInSeconds. The RFC 0008 variable table and Cancelation-behavior prose, plus the Template Schemas §4.3.1 table, now document the null case and the null-coalescing fallback pattern ({{ WrappedAction.Cancelation.Mode or 'TERMINATE' }}). The conformance fixture was renamed to wrap-cancelation-mode-null-when-no-cancelation and asserts both that null renders as the empty string in interpolation (MODE=<>) and that the nullness is observable via EXPR coalescing (COALESCED=<UNDECLARED>). Matching runtime changes landed in openjd-rs (44e31aa, seeds null and validates the symbol as string | null) and openjd-sessions-for-python (7198b5b); all 53 WRAP_ACTIONS conformance tests pass against both CLIs.
```

New conformance tests added.

seant-aws (approved):
should we have a few lines to concretely spell out plans for Windows/ non posix ?

from How-Jobs-Are-Run.md, and other places, we only mention `SIGKILL`,  as well I don't think there are any conformance tests speciically for windows. Just a consideration, not sure if okay for now? Doesn't have to be blocker but as a consideration

jericht (dismissed):

leongdl:
> should we have a few lines to concretely spell out plans for Windows/ non posix ?
> 
> from How-Jobs-Are-Run.md, and other places, we only mention `SIGKILL`, as well I don't think there are any conformance tests speciically for windows. Just a consideration, not sure if okay for now? Doesn't have to be blocker but as a consideration

Great question and I think we should build more tests for the future. Before the feature launch, lets try to bulk up all tests for all supported operating systems.

seant-aws (dismissed):

wiki/2023-09-Template-Schemas.md
```diff
@@ -1712,6 +1758,44 @@ The cancellation method defines the process by which an action is canceled.
...
+
+```yaml
+mode: "TERMINATE" | "NOTIFY_THEN_TERMINATE"
+mode: "TERMINATE" | "NOTIFY_THEN_TERMINATE" | <fmtstring> # @fmtstring @extension FEATURE_BUNDLE_1
+notifyPeriodInSeconds: <posinteger> | <posintstring> # @optional @fmtstring @extension FEATURE_BUNDLE_1
+```
+
+When *mode* is a format string:
+
+1. The value MUST consist of exactly one interpolation expression (`"{{ ... }}"`) with no surrounding
```
mwiebe: This doesn't sound right. If it's a string like "TERMINATE" that should work too.

EDIT: Realizing that "TERMINATE" is before format string in the grammar so that's not technically the format string case. Being this pedantic doesn't reduce expressiveness, but it feels odd that we'd prevent normal format string behavior.

mwiebe (dismissed):

AlexTranAmz (dismissed):

wiki/2023-09-Template-Schemas.md
leongdl: Good catch — fixed in 01b8903. The whole-field-only restriction is gone; a format-string `mode` now gets normal format string behavior:

- **Template Schemas §5.3**: the "MUST consist of exactly one interpolation expression" rule is replaced with normal resolution semantics — the resolved value must be `TERMINATE` or `NOTIFY_THEN_TERMINATE`, and the `string?` null case (dropping the whole `cancelation` object) applies only when the value happens to be a whole-field expression, mirroring the existing wording for `timeout` and `notifyPeriodInSeconds`.
- New conformance fixture `wrap-cancelation-partial-fmtstring-mode` pins the partial-interpolation case (`mode: "{{ 'NOTIFY' }}_THEN_TERMINATE"`).
- Matching runtime changes landed in openjd-rs (30cdf37), openjd-model-for-python (71fd0d3), and openjd-sessions-for-python (0295e21) — the static whole-field guards are removed, and null-vs-empty-string is distinguished so a non-whole-field mode resolving empty is an error rather than silently dropping the object. Full WRAP_ACTIONS conformance passes on both CLIs (Rust full suite: 1144/1144).

mwiebe (approved):

AlexTranAmz (approved):

seant-aws (approved):

PR #153, opened 2026-07-19
test: Add 13 WRAP_ACTIONS conformance fixtures from the coverage audit

leongdl:
## Why these conformance tests are needed

A five-pass deep review of the RFC 0008 wrap-actions feature (spec + openjd-rs + openjd-model-for-python + openjd-sessions-for-python) found that **every High-severity cross-runtime divergence lived on a spec-mandated path that no conformance fixture exercised**. The suite was passing 59/59 on both CLIs while the two implementations disagreed about single-layer enforcement, run-time cancelation validation, forwarded environment contents, environment-template scope, and the cleanup guarantee. The review's coverage audit walked every normative statement in Template Schemas §4.2/§4.3/§4.3.1/§5/§5.3, How-Jobs-Are-Run's wrap sections, and RFC 0008's Cancelation/Timeout/Cleanup/Stdout/Failure sections against the fixture list, element by element (positive / negative / edge), and produced the gap list this PR closes.

This PR adds **13 fixtures (59 → 72)**: 11 job execution tests and 2 env-template validation tests. Each fixture's header comment cites the exact spec sentence it pins.

### Fixture-by-fixture rationale

**Run-time cancelation validation — §5.3 steps 1–3 (3 fixtures).** Template Schemas §5.3 says a resolved format-string `mode` "MUST then validate against that form's schema" and "Any other resolved value is an error; the runtime MUST fail the action"; §5.3.2 caps `notifyPeriodInSeconds` at 600. None of this is statically checkable — expression values only exist at run time — and the suite had **zero** fixtures for it, which is exactly why one runtime shipped without launch-time enforcement (a typo'd forwarding expression is silently ignored and the wrapped task runs as if the author's cancelation semantics were honored):
- `wrap-cancelation-mode-resolves-invalid-fails` — mode resolves to `GARBAGE`; the action must fail and its script must never run (asserted via `forbidden:`).
- `wrap-cancelation-terminate-with-period-fails` — mode resolves `TERMINATE` with a non-null period; the resolved-shape validation must reject it. Parse time cannot catch this (a deferred mode accepts the union of both forms' fields by design).
- `wrap-cancelation-period-over-cap-fails` — `"{{ 700 }}"` bypasses the parse-time 600 cap; the runtime must enforce the bound at resolution.

**Round-trip forwarding through job instantiation (1 fixture).** All four existing round-trip fixtures place the wrap env in *external* environment templates, which never pass through job creation. `wrap-cancelation-roundtrip-job-environment` moves the identical forwarding pattern into `jobEnvironments`, so it routes through `create_job` — where `WrappedAction.*` cannot exist yet and implementations that eagerly resolve creation-scoped fields (`timeout`, `notifyPeriodInSeconds`) reject a valid template with "Undefined variable". This placement is spec-supported and is the follow-up spec's headline feature; the suite could not see the gap.

**Environment-template scope carried to wrap hooks (2 fixtures).** A reusable wrap template is parameterized (`parameterDefinitions`: container image, mount prefix — the RFC's own motivating examples) and computes derived values with script-level `let`. Both are part of the environment's format-string scope (§4 / §4.2), yet no fixture referenced either from a wrap hook — and the wrap-task hook resolves against a symbol table derived from the wrapped *step*, which knows nothing about the wrap env's own scope, so the runtime must carry it explicitly:
- `wrap-env-template-parameters-in-hooks` — `{{Param.Prefix}}` from the env template's own defaults, in all three hooks.
- `wrap-env-let-bindings-in-hooks` — `let greeting = ...` referenced from all three hooks.

**`WrappedAction.Environment` is openjd_env-only (1 fixture).** §4.3.1: the variable "carries only `openjd_env`-defined variables from the current session"; host-side variables are the wrapping environment's responsibility. `wrap-environment-excludes-variables-map` pins both directions with `output:` + `forbidden:`: an inner env's `openjd_env` export must appear, its declarative `variables:` entry must not. The review found the two runtimes handing different byte lists to the same wrap script here.

**Session-level single-layer rule (1 fixture).** How-Jobs-Are-Run: "If more than one Environment in the session stack defines any wrap hook, the session is invalid and the scheduler must reject it before entering any Environment." The existing `wrap-multiple-wrap-envs-rejected` stacks both wrap envs inside **one** job template — template validation catches that. `wrap-two-wrap-env-templates-rejected` supplies them as **separate** environment templates, where no single template is invalid and only the session/runner can enforce the rule. Without it, a runner that skips the session check silently runs only the innermost wrap env — the outer wrap layer's containment (potentially security-relevant) just doesn't happen, and the suite stays green.

**Cleanup guarantee (1 fixture).** RFC 0008's "Lifecycle and cleanup guarantees" scenario table had zero fixtures — previously hard to express, now possible with the runner's `taskFailure` exit-code assertion. `wrap-failed-enter-still-runs-wrap-exit` pins row 1: a failed `onWrapEnvEnter` (exit 7) must still run the inner env's `onWrapEnvExit` and the wrapping env's own `onExit`, surface exit code 7 as the failure, and never run the task. This is the "container leak" scenario: a runner that aborts without the exits orphans whatever the wrap env's `onEnter` started.

**Interception edge (1 fixture).** `onWrapEnvExit` "runs instead of the `onExit` action" — when the inner env defines no `onExit` there is nothing to replace, so the hook must not fire. `wrap-inner-env-without-on-exit-skips-exit-hook` pins the skip (both runtimes agree today; nothing pinned it). The enter-side twin is unrepresentable because the base schema requires `onEnter`.

**Macro propagation beyond `openjd_env` (1 fixture).** RFC 0008: runtimes MUST scan the wrap script's stdout for OpenJD macros, and verbatim forwarding makes a wrapped process's macro "recognized identically". Only `openjd_env` was pinned. `wrap-openjd-fail-from-wrapped-process` covers `openjd_fail` plus exit-status propagation (exit 3) through the wrap layer.

**Variable-scope rule beyond action args (2 env-template fixtures).** §4 items 5–7: schedulers must reject `WrappedAction.*` outside the three hooks. The one existing scope fixture covers action `args` on the job side. The review found scope enforcement regressing per-field on both implementations at different times — first one runtime missed cancelation fields, then the other — so these pin the two field classes that actually diverged:
- `4--wrappedaction-in-cancelation-outside-hook.invalid` — the env's own `onEnter` forwarding `cancelation.mode` (an easy copy-paste mistake from the round-trip pattern).
- `4--wrappedaction-in-embedded-file.invalid` — embedded file `data`; embedded files are shared by *all* the script's actions, including the env's own `onEnter`/`onExit` where `WrappedAction.*` never exists.

### Current implementation status (expected — these fixtures pin spec text, not current behavior)

Verified against both in-development CLIs before opening this PR:

| CLI | Result | Failing fixtures ↔ known open findings |
|---|---|---|
| openjd-rs (with OpenJobDescription/openjd-rs#265) | **69/72** | lets-in-hooks (F12), variables-map exclusion (F9), cleanup guarantee (F15) |
| openjd-sessions-for-python (RFC 0008 branch) | **68/72** | the three runtime cancelation-validation fixtures (F5) and the jobEnvironments round-trip (F2) |

Every failure maps one-to-one onto an already-tracked finding from the review, with fixes queued (Rust top-three already landed in openjd-rs#265; remaining Rust items and the Python items are next — we will endeavor to fix all major issues). The 6 fixtures where both implementations already agree are green on both. Per the conformance workflow's existing policy ("We expect some tests to fail with the openjd CLI at the moment"), red fixtures document spec/implementation gaps rather than blocking; as the queued fixes land, these flip green and become permanent regression protection.

The WRAP_ACTIONS README's layout and statement-to-fixture mapping are updated for all 13.

### DIFF
--- a/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-cancelation-period-over-cap-fails.test.yaml
+++ b/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-cancelation-period-over-cap-fails.test.yaml
@@ -1,0 +1,51 @@
+# Template Schemas 5.3.2: notifyPeriodInSeconds has a maximum value of
+# 600. A literal 700 is rejected at parse time, and a constant expression
+# like "{{ 700 }}" may equally be folded and rejected during validation —
+# so this fixture derives the out-of-range value from WrappedAction.*,
+# which is seeded per-action at host evaluation time and cannot be known
+# statically or at job creation: the wrapped action's forwarded period
+# (45) plus 655 resolves to 700 only when the hook is about to run. The
+# runtime must enforce the bound at resolution and fail the action;
+# otherwise the 600-second cap is bypassable by any run-time-derived
+# expression.
+template:
+  specificationVersion: jobtemplate-2023-09
+  name: WrapCancelationPeriodOverCapFails
+  steps:
+  - name: Step1
+    script:
+      actions:
+        onRun:
+          command: echo
+          args: ["placeholder"]
+          cancelation:
+            mode: NOTIFY_THEN_TERMINATE
+            notifyPeriodInSeconds: 45
+environments:
+- specificationVersion: environment-2023-09
+  extensions:
+  - WRAP_ACTIONS
+  - EXPR
+  - FEATURE_BUNDLE_1
+  environment:
+    name: WrapEnv
+    script:
+      actions:
+        onWrapEnvEnter:
+          command: bash
+          args: ["-c", "echo WRAP ENTER"]
+        onWrapTaskRun:
+          command: bash
+          args: ["-c", "echo PERIOD OVER CAP TASK RAN"]
+          cancelation:
+            mode: NOTIFY_THEN_TERMINATE
+            notifyPeriodInSeconds: "{{ WrappedAction.Cancelation.NotifyPeriodInSeconds + 655 }}"
+        onWrapEnvExit:
+          command: bash
+          args: ["-c", "echo WRAP EXIT"]
+runOn:
+- posix
+expected:
+  taskFailure: {}
+  forbidden:
+  - "PERIOD OVER CAP TASK RAN"
--- a/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-environment-excludes-variables-map.test.yaml
+++ b/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-environment-excludes-variables-map.test.yaml
@@ -1,0 +1,62 @@
+# Template Schemas 4.3.1: "WrappedAction.Environment carries only
+# openjd_env-defined variables from the current session". An inner
+# environment's declarative `variables:` map is applied to the real
+# subprocess environment by the runtime, but it is NOT an openjd_env
+# export and must not appear in the forwarded list — the wrapping
+# environment is responsible for host-side variables. This fixture pins
+# the exclusion: the inner env declares STATIC_VAR via `variables:` and
+# exports DYNAMIC_VAR via openjd_env (emitted by its wrapped onEnter,
+# forwarded verbatim by the wrap hook); only DYNAMIC_VAR may surface.
+template:
+  specificationVersion: jobtemplate-2023-09
+  name: WrapEnvironmentExcludesVariablesMap
+  jobEnvironments:
+  - name: InnerVars
+    variables:
+      STATIC_VAR: hello
+    script:
+      actions:
+        onEnter:
+          command: bash
+          args: ["-c", "echo 'openjd_env: DYNAMIC_VAR=world'"]
+        onExit:
+          command: bash
+          args: ["-c", "echo INNER EXIT BODY"]
+  steps:
+  - name: Step1
+    script:
+      actions:
+        onRun:
+          command: echo
+          args: ["placeholder"]
+environments:
+- specificationVersion: environment-2023-09
+  extensions:
+  - WRAP_ACTIONS
+  - EXPR
+  environment:
+    name: WrapEnv
+    script:
+      actions:
+        onWrapEnvEnter:
+          command: bash
+          args:
+          - "-c"
+          - "{{ repr_sh(WrappedAction.Command) }} {{ repr_sh(WrappedAction.Args) }}"
+        onWrapTaskRun:
+          command: bash
+          args:
+          - "-c"
+          - "printf 'ENVLINE=%s\\n' {{ repr_sh(WrappedAction.Environment) }}"
+        onWrapEnvExit:
+          command: bash
+          args:
+          - "-c"
+          - "{{ repr_sh(WrappedAction.Command) }} {{ repr_sh(WrappedAction.Args) }}"
+runOn:
+- posix
+expected:
+  output:
+  - "ENVLINE=DYNAMIC_VAR=world"
+  forbidden:
+  - "ENVLINE=STATIC_VAR"
--- a/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-inner-env-without-on-exit-skips-exit-hook.test.yaml
+++ b/conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-inner-env-without-on-exit-skips-exit-hook.test.yaml
@@ -1,11 +1,11 @@
-# Template Schemas 4.3 / How-Jobs-Are-Run: onWrapEnvExit "runs instead of
-# the onExit action of every inner environment". When the inner
-# environment defines no onExit there is nothing to replace, so the exit
-# hook must NOT fire for that environment — a runtime that runs the hook
-# unconditionally would execute wrapper teardown logic (e.g. a container
-# exec) against an action that does not exist. onEnter is required by the
-# base schema, so the enter side of this edge is unrepresentable; this
-# fixture pins the testable exit half.
+# Template Schemas 4.3, WRAP_ACTIONS constraint 5 (nothing-to-replace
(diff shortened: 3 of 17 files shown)


conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-environment-excludes-variables-map.test.yaml
```diff
@@ -0,0 +1,62 @@
+# Template Schemas 4.3.1: "WrappedAction.Environment carries only
+# openjd_env-defined variables from the current session". An inner
+# environment's declarative `variables:` map is applied to the real
+# subprocess environment by the runtime, but it is NOT an openjd_env
+# export and must not appear in the forwarded list — the wrapping
+# environment is responsible for host-side variables. This fixture pins
+# the exclusion: the inner env declares STATIC_VAR via `variables:` and
+# exports DYNAMIC_VAR via openjd_env (emitted by its wrapped onEnter,
+# forwarded verbatim by the wrap hook); only DYNAMIC_VAR may surface.
+template:
```
leongdl, line 10: A bit weiry about this one and would appreciate any comments.

leongdl: actually this makes sense, the inner enviornment variable declares the variable. The static one inside the inner environment isn't forwarded to the wrap.

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-inner-env-without-on-exit-skips-exit-hook.test.yaml
```diff
@@ -0,0 +1,52 @@
+# Template Schemas 4.3 / How-Jobs-Are-Run: onWrapEnvExit "runs instead of
+# the onExit action of every inner environment". When the inner
+# environment defines no onExit there is nothing to replace, so the exit
+# hook must NOT fire for that environment — a runtime that runs the hook
+# unconditionally would execute wrapper teardown logic (e.g. a container
+# exec) against an action that does not exist. onEnter is required by the
+# base schema, so the enter side of this edge is unrepresentable; this
+# fixture pins the testable exit half.
+template:
```
leongdl, line 9: This one seems controvertial, but makes sense given the spec.

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-cancelation-period-over-cap-fails.test.yaml
```diff
@@ -0,0 +1,44 @@
+# Template Schemas 5.3.2: notifyPeriodInSeconds has a maximum value of
+# 600. A literal 700 is rejected at parse time, but an expression value
+# ("{{ 700 }}", FEATURE_BUNDLE_1) cannot be checked statically — the
```
mwiebe: nit: That expression is a constant, it can be evaluated statically and resolved even during the first pass of template validation.

mwiebe: For both this and the `{{ 'GARBAGE' }}` case above, you could shift that into job creation time by making a job parameter. To make it require host evaluation time you'd have to add a variable dependency on something that is only available at that point in time, like an `if` expression that depends on the length of a path mapped value or something.

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-inner-env-without-on-exit-skips-exit-hook.test.yaml
```diff
@@ -0,0 +1,52 @@
+# Template Schemas 4.3 / How-Jobs-Are-Run: onWrapEnvExit "runs instead of
+# the onExit action of every inner environment". When the inner
+# environment defines no onExit there is nothing to replace, so the exit
+# hook must NOT fire for that environment — a runtime that runs the hook
```
crowecawcaw: Spec could benefit from explicitly stating this behavior. Right now it's implicit.

crowecawcaw (dismissed):

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-cancelation-period-over-cap-fails.test.yaml
leongdl: Good catch — fixed in 09a25fc. The comment claimed constant expressions can't be checked statically, which is wrong (they can be folded during validation). Reworked per your suggestion below so the value genuinely requires host evaluation time.

leongdl: Done in 09a25fc: all three runtime-validation fixtures now derive their invalid values from WrappedAction.* (forwarded period + 655 → 700 over the cap; forwarded mode + "_GARBAGE" suffix → non-mode string; forwarded TERMINATE + literal period → shape violation), which is seeded per-action at host evaluation time and can't be constant-folded or resolved at job creation. Verified the expected pass/fail matrix is unchanged (Rust 69/72, Python 68/72 — the reds map to the known F5/F2/F9/F12/F15 findings).

conformance-tests/2023-09/WRAP_ACTIONS/jobs/wrap-inner-env-without-on-exit-skips-exit-hook.test.yaml
leongdl: Updated the spec with a sentence.

crowecawcaw (approved):

PR #154, opened 2026-07-22
fix: Make WRAP_ACTIONS fixtures cross-platform via python; include `variables:` in `WrappedAction.Environment`

mwiebe:
This PR does two related things for the `WRAP_ACTIONS` extension (RFC 0008):
it lets more of its conformance tests run on Windows, and it fixes a gap in
what the spec says a wrap script gets to see.

### Background

`WRAP_ACTIONS` lets an environment "wrap" the commands a job would normally
run — instead of running the task directly, the runtime calls the wrap
environment's hook script and hands it a description of the wrapped command
through `WrappedAction.*` values (`.Command`, `.Args`, `.Environment`, ...).
The classic use case is a reusable wrapper that launches every task inside a
container: the wrap script reads those values and re-runs the command inside
`docker run`.

The conformance suite is a set of small job templates ("fixtures") that any
OpenJD implementation can run to check it behaves the way the spec says.
Each fixture runs some tiny commands and asserts on their output. A fixture
can carry a `runOn:` gate to say "only run me on these platforms".

### Change 1: fixtures run on Windows now

Nine WRAP_ACTIONS fixtures used `bash` or bare `echo` for their tiny
commands. Since those aren't reliably available on Windows, seven of them
were gated to `runOn: [posix]` — meaning Windows implementations simply
never exercised the behaviors they pin (environment-template parameters and
`let` bindings inside wrap hooks, cleanup ordering after a failed enter, and
several validation rules).

But none of those fixtures actually test anything shell-specific; they just
need *some* command that prints a line. The rest of the conformance suite
already has a convention for that: `python` (used by ~935 fixtures — it's
the suite's portable interpreter, guaranteed on test hosts on every
platform). This PR switches the nine fixtures to `python -c "print(...)"`
one-liners, drops their `runOn: [posix]` gates, and documents the
convention in the WRAP_ACTIONS README: fixtures with no gate use python;
genuinely shell-specific fixtures stay gated.

### Change 2: `WrappedAction.Environment` includes `variables:` maps

There are two ways a job defines environment variables for its processes:

1. A dynamic one: an action prints `openjd_env: NAME=value` and the runtime
   picks it up.
2. A declarative one: an environment lists them in a `variables:` map in
   the template.

To the wrapped process these are indistinguishable — the runtime applies
both to its environment. But the spec said `WrappedAction.Environment` (the
list the wrap script receives) carries only the first kind. That means a
container wrapper forwarding "the session's variables" into the container
would silently drop everything declared via `variables:` — the wrapped
process behaves differently inside the wrapper than outside it, which is
exactly what wrapping is supposed to avoid.

This PR changes the spec so `WrappedAction.Environment` carries both:
everything the session defined, however it was declared. (Host-inherited
variables like `PATH` are still excluded — forwarding those is the
wrapper's own business.) Updated in RFC 0008 and the 2023-09 Template
Schemas wiki, and the fixture that pinned the old exclusion is renamed
`wrap-environment-excludes-variables-map` →
`wrap-environment-includes-variables-map` with its assertions inverted:
the declared variable must now appear in the forwarded list instead of
being forbidden.

---


conformance-tests/2023-09/WRAP_ACTIONS/README.md
```diff
@@ -277,9 +278,10 @@ bundles in this directory:
   `wrap-env-let-bindings-in-hooks`) — the wrapped step's symbol table
   knows nothing about them, so the runtime must carry the environment's
   own scope to its hooks.
-- **`WrappedAction.Environment` is openjd_env-only** (§4.3.1): an inner
```
leongdl, line 280: Yikes, this is 100% bug.

rfcs/0008-environment-wrap-actions.md
```diff
@@ -306,7 +306,7 @@ helper scripts can be reused unchanged.
 |-----------------------------------|----------------|-------------|
 | `WrappedAction.Command`           | `string`       | The `command` from the wrapped action. |
 | `WrappedAction.Args`              | `list[string]` | The `args` from the wrapped action. |
-| `WrappedAction.Environment`       | `list[string]` | Environment variables defined by `openjd_env` earlier in the session, as `["KEY=value", ...]`. See [Host environment variables and embedded file paths](#host-environment-variables-and-embedded-file-paths). |
+| `WrappedAction.Environment`       | `list[string]` | Environment variables defined earlier in the session — `openjd_env` exports and environments' declarative `variables:` maps — as `["KEY=value", ...]`. See [Host environment variables and embedded file paths](#host-environment-variables-and-embedded-file-paths). |
```
leongdl, line 309: Great catch!

leongdl (approved):

PR #162, opened 2026-08-02
test: repr_py control-character conformance tests

leongdl:
Two fixtures pinning Expression Language §2.2.6's requirement that `repr_py` escape control characters, promoted out of `proposed/` now that the fix has landed.

**TLDR.** Both were parked here as expected failures. [openjd-rs#374](https://github.com/OpenJobDescription/openjd-rs/pull/374) merged as `5b04959`, so both pass and become the regression tests the `proposed/README.md` promised. Verified against a pre-fix binary as the control: both fail before the fix and pass after, so they pin what they claim.

## The two fixtures

| Fixture | What it pins |
|---|---|
| `EXPR/jobs/expr2.2.6--repr-py-newline-roundtrip` | `repr_py("a\nb")` fed to `ast.literal_eval` round-trips to the input |
| `WRAP_ACTIONS/jobs/wrap-repr-py-escapes-newline-in-wrapped-args` | the same rule through RFC 0008's `repr_py(WrappedAction.Args)` forwarding pattern, with a two-line `python -c` body |

The pair is deliberate. The EXPR one isolates the rule; the WRAP_ACTIONS one exercises the end-to-end path a template author actually writes, where the value crosses two process boundaries before being parsed.

## Measured

Built `openjd` from openjd-rs `main` at `5b04959`, and separately from a pre-fix commit as the control:

| Fixture | pre-fix | merged |
|---|---|---|
| `expr2.2.6--repr-py-newline-roundtrip` | FAIL | **PASS** |
| `wrap-repr-py-escapes-newline-in-wrapped-args` | FAIL | **PASS** |

Both pre-fix failures are `SyntaxError: unterminated string literal` with the expected output missing, which is exactly the defect: `repr_py` quoted its output without escaping, so a value carrying a newline emitted it raw and CPython would not parse the result.

Run through the runner's **directory scan** from the new locations, not single-file mode, since discoverability is the point of leaving `proposed/`:

```
2023-09/EXPR/jobs:
  ✓ expr2.2.6--repr-py-newline-roundtrip
2023-09/WRAP_ACTIONS/jobs:
  ✓ wrap-repr-py-escapes-newline-in-wrapped-args
```

No regression in either category: **WRAP_ACTIONS 73 passed, 0 failed** and **EXPR 352 passed, 0 failed**.

## Notes for review

The headers are rewritten rather than moved unchanged. Both described the parked state and asserted that the implementations fail, which is no longer true. They now carry spec, test and expectation in three paragraphs, name the paired fixture, and sit inside the comment-line budget.

`proposed/README.md` and the `proposed/` directory are removed. The fixture it described was its only entry.

The EXPR fixture originates on the func-lib expected-failures branch behind [#165](https://github.com/OpenJobDescription/openjd-specifications/pull/165), which should drop its copy so the two do not collide on merge.

One thing this does **not** resolve, recorded so it is not lost: Template Schemas §5.2 restricts `<ArgString>` to characters outside the `Cc` category, which excludes the newlines both of these fixtures rely on. Neither implementation enforces that, and 150 of 1,191 merged fixtures carry a newline in an `args` entry, so the multi-line `python -c` convention is load-bearing suite-wide. Either §5.2 is relaxed to permit `U+000A` and `U+000D`, or that convention is invalid. That is a spec decision and a separate PR; these fixtures are correct under the reading both implementations already follow.

----

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*


conformance-tests/2023-09/WRAP_ACTIONS/jobs/proposed/wrap-repr-py-escapes-newline-in-wrapped-args.test.yaml
leongdl, line 1: **Quorum verdict: GOOD-WITH-NITS (4 reviewers flagged variants of the same two issues)** — spec: EL §2.2.6 repr_py (newline must be escaped — spec gives the exact worked example) via RFC 0008 WrappedAction.Args forwarding.

(1) Spec self-tension, correctly disclosed but under-weighted: the fixture's own onRun args contain a literal U+000A, which §5.2's ArgString Cc-exclusion arguably forbids — as written the template may be spec-invalid, so promotion is contingent on resolving §5.2 vs the multi-line `python -c` convention (which the Deadline Docker container env also relies on; if §5.2 is enforced, that breaks service-wide). Resolve the spec question first.
(2) Attribution: README says "implementation bug (openjd-rs)" but the 2026-08-12 sweep shows it fails BOTH implementations — state Python's status.
(3) Partial overlap with PR #165's expr2.2.6--repr-py-newline-roundtrip (same root defect); the end-to-end wrap path justifies keeping both, but cross-reference them.

leongdl:
Quorum review (5 independent agents). 1 fixture + README. Verdict: GOOD-WITH-NITS. The repr_py newline defect is real (fails BOTH implementations per the 2026-08-12 sweep) and the end-to-end wrap-forwarding angle adds value over PR #165's unit-level twin — but the fixture is hostage to an unresolved spec tension its own README flags, and the README under-attributes the failure. Cross-PR: this branch uses component-level WRAP_ACTIONS/proposed/ while PR #166 uses kind-level job_templates/proposed/ — the five expected-failures PRs should standardize on one placement (kind-level routes promotion mechanically).

leongdl:

- Restructured to kind-level parking: `WRAP_ACTIONS/jobs/proposed/` (promotion = move up one directory).
- README corrected: the repr_py newline defect fails **both** implementations (2026-08-12 sweep), not just openjd-rs.
- The §5.2 ArgString/newline conflict is now the stated **promotion gate** rather than a footnote — as written the fixture's own args are arguably spec-invalid until that's resolved (and the resolution also decides the fate of the multi-line `python -c` convention suite-wide).
- Cross-referenced the unit-level twin in the func-lib expected-failures PR (#165).

conformance-tests/2023-09/WRAP_ACTIONS/jobs/proposed/README.md
```diff
@@ -0,0 +1,47 @@
+# Proposed WRAP_ACTIONS fixtures (parked) — jobs
```
leongdl: WIP - have a fix pending and proposed.

AlexTranAmz (approved):

conformance-tests/2023-09/WRAP_ACTIONS/jobs/proposed/README.md
leongdl: Fix released.

Issue #107, opened 2026-02-10
Clarify §1.2.1: Constraint merging uses intersection semantics


crowecawcaw:
Section 1.2.1 states:

> These constraints do not need to be identical in every Job Parameter Definition for the same parameter name, but each constraint must become more constrained as a subsequent 
definition is merged with the previous ones in processing order

The phrase "must become more constrained" is ambiguous. It could be read as requiring each definition to have stricter constraints than the previous, which would make wider constraints 
an error.

The intended behavior is intersection semantics: the effective constraint is the most restrictive combination of all definitions. A wider constraint in a later definition is allowed but
has no effect.

Possible rewording:

> These constraints do not need to be identical in every Job Parameter Definition for the same parameter name. The effective constraint for each property is the most 
restrictive combination of all definitions. For example, if one definition has minLength: 3 and another has minLength: 5, the effective minimum length is 5.

PR #163, opened 2026-08-02
[WIP] test: Promote base int64 and merged-default fixtures into the live suite

leongdl:
> [!WARNING]
> **The conformance workflows go red on this PR, deliberately.** These fixtures were parked in
> `base/*/proposed/`, which the runner does not scan, so nothing ran them. They are now live so the
> failures are visible. Six of the twelve are gated on a spec erratum that does not exist yet and
> should not merge before it lands. See "Do not merge as-is" below.

12 fixtures across two families, all `.invalid`. Measured over the full `2023-09/base` suite against
openjd-cli 0.7.7 and openjd-rs `main` e4efd67: Rust 665 passed / 9 failed, Python 668 passed /
6 failed. The base suite has no other failures on either implementation, so every failure belongs to
this PR.

| Fixture | Python | Rust |
|---|---|---|
| `job_templates/2--type-lowercase.invalid.yaml` | pass | pass |
| `job_templates/2.3--int-default-above-int64-max.invalid.yaml` | FAIL | FAIL |
| `job_templates/2.3--int-default-below-int64-min.invalid.yaml` | FAIL | FAIL |
| `job_templates/2.3--int-minvalue-above-int64-max.invalid.yaml` | FAIL | FAIL |
| `job_templates/2.3--int-maxvalue-above-int64-max.invalid.yaml` | FAIL | FAIL |
| `job_templates/2.3--int-allowedvalues-above-int64-max.invalid.yaml` | FAIL | FAIL |
| `job_templates/2.3--int-default-above-int64-max-intstring.invalid.yaml` | FAIL | pass |
| `jobs/1.2.1--merged-default-below-merged-minvalue.invalid.test.yaml` | pass | FAIL |
| `jobs/1.2.1--merged-default-below-merged-minlength.invalid.test.yaml` | pass | FAIL |
| `jobs/1.2.1--merged-default-not-in-merged-allowedvalues-int.invalid.test.yaml` | pass | FAIL |
| `jobs/1.2.1--merged-default-below-merged-minvalue-float.invalid.test.yaml` | pass | FAIL |
| `jobs/2.3--int-value-above-int64-max.invalid.test.yaml` | pass | pass |

## A merged default is not checked against the merged constraint (4 fixtures)

Section 1.2.1 requires the merged job parameter definition to be internally consistent and names the
`default` as its example. Each fixture gives the job template, processed last, a default that
satisfies its own bound but violates the bound merged in from an environment template. Python
refuses all four at job generation. openjd-rs runs them.

Rejection follows under both readings of "each constraint must become more constrained": the literal
one, and the intersection reading proposed in #107. These fixtures do not depend on #107 closing.

Each was probed with a control that both implementations accept, so each isolates one variable:
`default: 50` inside the merged `[10, 100]`, `default: abcdef` at the merged minimum 5, `default: 3`
inside the merged `[1, 2, 3]`, `default: 15.0` above the merged 10.0. The supplied-value equivalents
(`-p Count=5`, `-p Value=abc`) are refused by both, which is what leaves the default uncovered.

Three of the four axes are separate code paths in openjd-rs, not one. `check_constraints` has
separate INT, FLOAT and STRING arms, and the default-vs-`allowedValues` check in
`validate_satisfiable` covers STRING and PATH only. That is why the live
`1.2.1--default-must-satisfy-merged-constraints.invalid.test.yaml` (STRING `allowedValues`) already
passes while the INT `allowedValues` fixture here does not. PATH is deliberately left uncovered:
openjd-rs joins a PATH default to the template directory before checking, so the outcome would
depend on the runner's working directory.

## INT and the int64 boundary (7 fixtures)

Both implementations accept `default`, `minValue`, `maxValue` and `allowedValues` at 2^63 and
`default` at -2^63-1.

What the acceptance costs, measured: openjd-rs accepts `default: 9223372036854775808` at
`openjd check` and then resolves the default to `9223372036854775807` at run time. Silent
saturation, not an error. The Python CLI accepts the same template at `check` and dies at `run` with
`Cannot convert '9223372036854775808' to int: number too large to fit in target type`, raised from
`_serialize_symbol_table` in `_create_job.py`. An uncaught `ValueError`, not validation.

`jobs/2.3--int-value-above-int64-max.invalid.test.yaml` passes on both for that reason, and Python
passes it by accident. Probed directly: `preprocess_job_parameters` accepts the value, `create_job`
accepts it, and only `create_job_with_symbol_tables` rejects. An earlier revision of this branch
claimed the Python CLI accepts it; that is true of `create_job` and false of the CLI.

The `<intstring>` fixture is a falsifiability guard. openjd-rs parses the string form with
`str::parse::<i64>` and already rejects it. Python coerces both forms to an unbounded `int`, so a
fix in `_precheck_is_int_type` that only checks `isinstance(value, int)` would pass all five
bare-integer fixtures and still accept the string.

## Type-name case (1 fixture, passes both)

`2--type-lowercase.invalid.yaml` was parked as an openjd-rs bug. It is not one any more: #364 gated
type-name case on the EXPR extension, and that fix ships in the published `openjd-cli` 0.1.18 the
Rust lane installs. Its positive twin is `EXPR/job_templates/2--type-case-insensitive.yaml`.

## Do not merge as-is

The six int64 fixtures assert a rule the base spec does not state. Section 2.3 types these
properties as `<integer>` and sets no bound; the int64 range appears only in the Expression Language
type table, an EXPR document. A bignum implementation is arguably conformant as the text stands.
They need a section 2.3 integer-bounds erratum first, and no issue for it exists yet. The
saturation measurement above is the case for filing one.

The four merged-default fixtures and `2--type-lowercase` are ready. Splitting them out of this PR is
the fastest way to get a green base suite with the merge gap recorded.

## Two runner limitations found while measuring

Neither is a defect in these fixtures. `run_job` returns `result.returncode == 0` for an `.invalid.`
fixture before it reads `expected`, so a rejection for the wrong reason, or a crash, counts as a
pass. That is how the Python traceback above went unnoticed, and it means adding `expected` entries
to an `.invalid.` fixture has no effect. Separately, a positive `.test.yaml` with no `expected` block
falls through to `return True` without consulting `returncode`, so it passes even when the CLI exits
non-zero.

## Changes in this revision

Every header rewritten to spec, test, expectation. The merged-default headers no longer narrate
openjd-rs internals or point at twin fixtures that live on an unmerged branch, so they stay accurate
once a fix lands. Both `proposed/README.md` inventories deleted; their measurement data is in this
description. Four fixtures added. Non-ASCII characters removed from every fixture, since the runner
opens them at the locale encoding and a UTF-8 character is mojibake on Windows.

### DIFF
--- a/conformance-tests/2023-09/base/job_templates/proposed/2--type-lowercase.invalid.yaml
+++ b/conformance-tests/2023-09/base/job_templates/proposed/2--type-lowercase.invalid.yaml
@@ -1,0 +1,15 @@
+# Parameter type names are case-sensitive in base (case-insensitivity is EXPR-only); 'string' must be rejected.
+specificationVersion: jobtemplate-2023-09
+name: TestJob
+parameterDefinitions:
+- name: MyParam
+  type: string
+steps:
+- name: Step1
+  script:
+    actions:
+      onRun:
+        command: python
+        args:
+        - "-c"
+        - "print()"
--- a/conformance-tests/2023-09/base/job_templates/proposed/2.3--int-allowedvalues-above-int64-max.invalid.yaml
+++ b/conformance-tests/2023-09/base/job_templates/proposed/2.3--int-allowedvalues-above-int64-max.invalid.yaml
@@ -1,0 +1,17 @@
+# An INT parameter allowedValues entry of 2^63 exceeds the signed 64-bit integer range and must be rejected.
+specificationVersion: jobtemplate-2023-09
+name: TestJob
+parameterDefinitions:
+- name: MyParam
+  type: INT
+  allowedValues:
+  - 9223372036854775808
+steps:
+- name: Step1
+  script:
+    actions:
+      onRun:
+        command: python
+        args:
+        - "-c"
+        - "print('{{Param.MyParam}}')"
--- a/conformance-tests/2023-09/base/job_templates/proposed/2.3--int-default-above-int64-max.invalid.yaml
+++ b/conformance-tests/2023-09/base/job_templates/proposed/2.3--int-default-above-int64-max.invalid.yaml
@@ -1,0 +1,16 @@
+# An INT parameter default of 2^63 exceeds the signed 64-bit integer range and must be rejected.
+specificationVersion: jobtemplate-2023-09
+name: TestJob
+parameterDefinitions:
+- name: MyParam
+  type: INT
+  default: 9223372036854775808
+steps:
+- name: Step1
+  script:
+    actions:
+      onRun:
+        command: python
+        args:
+        - "-c"
+        - "print('{{Param.MyParam}}')"
--- a/conformance-tests/2023-09/base/job_templates/proposed/2.3--int-default-below-int64-min.invalid.yaml
+++ b/conformance-tests/2023-09/base/job_templates/proposed/2.3--int-default-below-int64-min.invalid.yaml
@@ -1,0 +1,16 @@
+# An INT parameter default of -2^63-1 is below the signed 64-bit integer range and must be rejected.
+specificationVersion: jobtemplate-2023-09
+name: TestJob
+parameterDefinitions:
+- name: MyParam
+  type: INT
+  default: -9223372036854775809
+steps:
+- name: Step1
+  script:
+    actions:
+      onRun:
+        command: python
+        args:
+        - "-c"
+        - "print('{{Param.MyParam}}')"
--- a/conformance-tests/2023-09/base/job_templates/proposed/2.3--int-maxvalue-above-int64-max.invalid.yaml
+++ b/conformance-tests/2023-09/base/job_templates/proposed/2.3--int-maxvalue-above-int64-max.invalid.yaml
@@ -1,0 +1,16 @@
+# An INT parameter maxValue of 2^63 exceeds the signed 64-bit integer range and must be rejected.
+specificationVersion: jobtemplate-2023-09
+name: TestJob
+parameterDefinitions:
+- name: MyParam
+  type: INT
+  maxValue: 9223372036854775808
+steps:
+- name: Step1
+  script:
+    actions:
+      onRun:
+        command: python
+        args:
+        - "-c"
+        - "print('{{Param.MyParam}}')"
--- a/conformance-tests/2023-09/base/job_templates/proposed/2.3--int-minvalue-above-int64-max.invalid.yaml
+++ b/conformance-tests/2023-09/base/job_templates/proposed/2.3--int-minvalue-above-int64-max.invalid.yaml
@@ -1,0 +1,16 @@
+# An INT parameter minValue of 2^63 exceeds the signed 64-bit integer range and must be rejected.
+specificationVersion: jobtemplate-2023-09
+name: TestJob
+parameterDefinitions:
+- name: MyParam
+  type: INT
+  minValue: 9223372036854775808
+steps:
+- name: Step1
+  script:
+    actions:
+      onRun:
+        command: python
+        args:
+        - "-c"
+        - "print('{{Param.MyParam}}')"
--- a/conformance-tests/2023-09/base/jobs/proposed/1.2.1--constraint-widening-int-range.invalid.test.yaml
+++ b/conformance-tests/2023-09/base/jobs/proposed/1.2.1--constraint-widening-int-range.invalid.test.yaml
@@ -1,0 +1,32 @@
+# Spec 1.2.1: each constraint must become MORE constrained in template processing order.
+# The job template (processed last) widens the environment template's INT range, so the
+# merge must be rejected. The default satisfies both ranges, isolating the widening rule.
+template:
+  specificationVersion: jobtemplate-2023-09
+  name: TestJob
+  parameterDefinitions:
+  - name: Count
+    type: INT
+    minValue: 0
+    maxValue: 1000
+    default: 50
+  steps:
+  - name: Step1
+    script:
+      actions:
+        onRun:
+          command: python
+          args:
+          - -c
+          - print(r'Count={{Param.Count}}')
+environments:
+- specificationVersion: environment-2023-09
+  parameterDefinitions:
+  - name: Count
+    type: INT
+    minValue: 10
+    maxValue: 100
+  environment:
+    name: TestEnv
+    variables:
+      COUNT: '{{Param.Count}}'
--- a/conformance-tests/2023-09/base/jobs/proposed/1.2.1--constraint-widening-minlength.invalid.test.yaml
+++ b/conformance-tests/2023-09/base/jobs/proposed/1.2.1--constraint-widening-minlength.invalid.test.yaml
@@ -1,0 +1,30 @@
+# Spec 1.2.1: each constraint must become MORE constrained in template processing order.
+# The job template (processed last) relaxes the environment template's minLength, so the
+# merge must be rejected. The default satisfies both, isolating the widening rule.
+template:
+  specificationVersion: jobtemplate-2023-09
+  name: TestJob
+  parameterDefinitions:
+  - name: Value
+    type: STRING
+    minLength: 2
+    default: test-value
+  steps:
+  - name: Step1
+    script:
+      actions:
+        onRun:
+          command: python
+          args:
+          - -c
+          - print(r'Value={{Param.Value}}')
+environments:
+- specificationVersion: environment-2023-09
+  parameterDefinitions:
+  - name: Value
+    type: STRING
+    minLength: 5
+  environment:
+    name: TestEnv
+    variables:
+      VAL: '{{Param.Value}}'
--- a/conformance-tests/2023-09/base/job_templates/proposed/README.md
+++ b/conformance-tests/2023-09/base/job_templates/proposed/README.md
@@ -1,0 +1,57 @@
+# Proposed conformance fixtures (expected failures) — base/job_templates
+
+Every fixture in this directory is parked because it **fails against at least
+one current reference implementation**. Placement is kind-level
+(`<component>/<kind>/proposed/`) so promotion is a mechanical move up one
+directory once the underlying issue is resolved. The conformance runner does
+not scan `proposed/` directories.
+
+Verification environment: `openjd` CLI = openjd-rs release build from
+upstream/main (post range-cap fix); cross-checks with openjd-model-for-python
+0.11.x via `decode_job_template`.
+
+## INT 64-bit boundary rejects
+
+Spec §2.3 types these fields as `<integer>` with **no stated bound** — the
+base 2023-09 document never mentions 64-bit anywhere; the int64 range exists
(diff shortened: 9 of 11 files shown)


conformance-tests/2023-09/base/jobs/proposed/1.2.1--constraint-widening-int-range.invalid.test.yaml
leongdl, line 1: **Quorum verdict: GOOD-WITH-NITS (5/5 with the same caveat)** — spec: §1.2.1 ("each constraint must become more constrained", Job Template processed last). The literal reading supports rejection and the README honestly offers the alternative (intersection-merge) reading — this is closer to a spec decision than a clean mandate; the fixture may be invalidated by a spec revision rather than an implementation fix. Service note: adoption flips CreateJob accept→reject for queue-environment parameter merging (BealineDataPlane grafts queue-env param definitions onto customer templates), and the un-scoped constraint-OMISSION case is the real-world hazard — gate promotion on the spec decision plus a service deprecation path. Matches sweep (fails both). Accept twin: mainline 1.2.1--constraint-narrowing-int-range.

conformance-tests/2023-09/base/jobs/proposed/1.2.1--constraint-widening-minlength.invalid.test.yaml
leongdl, line 1: **Quorum verdict: GOOD-WITH-NITS** — spec: §1.2.1, STRING minLength axis. Same narrow-vs-intersection ambiguity and same queue-env merge hazard as the int-range twin; good isolation (default satisfies both ranges). Accept twin: mainline 1.2.1--constraint-narrowing-minlength.

conformance-tests/2023-09/base/job_templates/proposed/2--type-lowercase.invalid.yaml
leongdl, line 1: **Quorum verdict: GOOD (5/5, one dedupe action)** — spec: §2 (base type names case-sensitive; case-insensitivity is EXPR-gated per RFC 0007 — the BNF literal plus the §2 sentence's negative implication is solid). Matches sweep (openjd-rs accepts, Python rejects) and the README states both. Action: this duplicates PR #166's base/job_templates/proposed/2--type-lowercase-string.invalid.yaml — same construct, same defect, two paths; keep one (or differentiate on the type token).

conformance-tests/2023-09/base/job_templates/proposed/2.3--int-default-above-int64-max.invalid.yaml
leongdl, line 1: **Quorum verdict: GOOD-WITH-NITS (majority; 1 NOT-GOOD dissent)** — the defect is real (both implementations accept 2^63; matches sweep) and rejection is a safe data-integrity adoption (values that can't round-trip int64 storage). Dissent (spec-literalist): the base 2023-09 spec never bounds <integer> — int64 exists only in the EXPR doc — so a bignum implementation is arguably conformant and this is a SPEC QUESTION, not an implementation bug. Consensus action: keep parked, but reword the README classification to 'pending base-spec integer-bounds erratum' and make that erratum the promotion gate. Also: README cites accept twin jobs/2.3--int-value-above-int64-max.invalid.test.yaml as 'in the main suite' — it's on PR #161's branch (and per that PR's review, it should move HERE).

conformance-tests/2023-09/base/job_templates/proposed/2.3--int-default-below-int64-min.invalid.yaml
leongdl, line 1: **Quorum verdict: GOOD-WITH-NITS** — same family, negative bound (-2^63-1). Matches sweep; accept twin 2.3--int-default-int64-min.yaml in PR #161. Same spec-erratum caveat as the family.

conformance-tests/2023-09/base/job_templates/proposed/2.3--int-minvalue-above-int64-max.invalid.yaml
leongdl, line 1: **Quorum verdict: GOOD-WITH-NITS** — same family, minValue axis; per-field decomposition justified (separate parse paths). Matches sweep; twin in PR #161. Inventory note: negative-side rejects (minValue/maxValue/allowedValues below -2^63) are uncovered — likely same code path, but note it.

conformance-tests/2023-09/base/job_templates/proposed/2.3--int-maxvalue-above-int64-max.invalid.yaml
leongdl, line 1: **Quorum verdict: GOOD-WITH-NITS** — same family, maxValue axis. Service note: likeliest of the five to hit real templates (maxValue=2^63 as an infinity sentinel) — flag in any service change notice. Matches sweep; twin in PR #161.

conformance-tests/2023-09/base/job_templates/proposed/2.3--int-allowedvalues-above-int64-max.invalid.yaml
leongdl, line 1: **Quorum verdict: GOOD-WITH-NITS** — same family, allowedValues axis; an unrepresentable allowedValue is dead weight, low blast radius. Matches sweep; twin in PR #161.

leongdl:
Quorum review (5 independent agents). 8 fixtures + README. Net: 2 clean GOOD (type-lowercase, and the widening pair is GOOD-WITH-NITS), 6 GOOD-WITH-NITS. The central contested point: the five int64 fixtures document behavior BOTH implementations exhibit (matches the sweep) and are safe data-integrity adoptions — but the base 2023-09 spec text never bounds <integer> (int64 lives only in the EXPR doc), so the spec-literalist reviewer voted all five NOT-GOOD as 'spec question misclassified as implementation bug'. Majority keeps them parked here, with the classification updated and a base-spec integer-bounds erratum filed as the promotion gate. Also: 2--type-lowercase duplicates PR #166's 2--type-lowercase-string (same construct, same defect) — dedupe across the two PRs. README quality is the best of the five expected-failures branches (genuine dual-implementation attribution).

leongdl:

- Restructured to kind-level parking: 6 template fixtures → `base/job_templates/proposed/`, the 2 widening fixtures → `base/jobs/proposed/` (promotion = move up one directory); README split per kind.
- **Adopted** `2.3--int-value-above-int64-max.invalid.test.yaml` from the base-gaps PR (#161): openjd-rs rejects the supplied 2^63 but openjd-model 0.11.x `create_job` accepts it — both verified by execution this session — so it belongs in the inventory, not the live suite. (The old README claim that it "lives in the main suite" and is "rejected correctly" was true only of openjd-rs.)
- int64 family reclassified: **spec question first** — the base 2023-09 doc never bounds `<integer>` (int64 exists only in the EXPR type table), so promotion is gated on a base-spec integer-bounds erratum; a bignum implementation is arguably conformant as the text stands.
- Widening entries now note the scheduler-merge blast radius (queue-environment parameter grafting) as part of the spec-decision gate.
- Noted the duplicate type-lowercase pin in #166 was dropped; this branch's copy is the single source.

conformance-tests/2023-09/base/jobs/proposed/1.2.1--constraint-widening-int-range.invalid.test.yaml
```diff
@@ -0,0 +1,32 @@
+# Spec 1.2.1: each constraint must become MORE constrained in template processing order.
+# The job template (processed last) widens the environment template's INT range, so the
+# merge must be rejected. The default satisfies both ranges, isolating the widening rule.
```
mwiebe, line 3: This sounds wrong, the result of the merge should be the intersection and give minValue 10, maxValue 100, not a failure.

> but each constraint must become more constrained as a subsequent definition is merged with the previous ones in processing order

This means that the resulting constraint we produce becomes more constrained. It does not mean that each subsequent constraint we take must be more constrained.

mwiebe (changes requested):

conformance-tests/2023-09/base/jobs/proposed/1.2.1--constraint-widening-int-range.invalid.test.yaml
leongdl: Thanks! I will adjust the test case, and then get fixes up.

conformance-tests/2023-09/base/jobs/proposed/1.2.1--constraint-widening-int-range.invalid.test.yaml
leongdl: Addressed with a8d721b. The maintainer review confirmed the intersection reading, so this fixture was removed rather than promoted. Replaced by `1.2.1--merged-default-below-merged-minvalue.invalid.test.yaml`.

conformance-tests/2023-09/base/jobs/proposed/1.2.1--constraint-widening-minlength.invalid.test.yaml
leongdl: Addressed with a8d721b. Same disposition as the int-range twin: fixture removed, replaced by `1.2.1--merged-default-below-merged-minlength.invalid.test.yaml`.

conformance-tests/2023-09/base/job_templates/proposed/2--type-lowercase.invalid.yaml
leongdl: Addressed with f5837f2. The duplicate in #166 was dropped; this branch's `2--type-lowercase.invalid.yaml` is the single copy (README notes the dedupe).

conformance-tests/2023-09/base/job_templates/proposed/2.3--int-default-above-int64-max.invalid.yaml
leongdl: Addressed with f5837f2. README reclassifies the int64 family as spec-question-first, gated on a base-spec integer-bounds erratum. `2.3--int-value-above-int64-max.invalid.test.yaml` moved here from #161 into `base/jobs/proposed/`, with the README corrected to say openjd-rs rejects and openjd-model 0.11.x accepts.

conformance-tests/2023-09/base/jobs/proposed/1.2.1--constraint-widening-int-range.invalid.test.yaml
leongdl: Addressed with a8d721b. You are right: §1.2.1 constrains the merged result, and issue #107 records the same intersection intent. Both `1.2.1--constraint-widening-*` fixtures are deleted. Their replacements, `1.2.1--merged-default-below-merged-minvalue` and `-minlength`, keep the same template shape but assert the last sentence of §1.2.1: the merged definition's `default` must satisfy the merged (intersected) constraints. Python refuses those jobs; openjd-rs runs them (`preprocess_job_parameters` only calls `check_constraints` on supplied values), so they stay parked as expected failures.

PR #164, opened 2026-08-02
test: [expected failures] EXPR fixtures needing implementation fixes

leongdl:
> [!IMPORTANT]
> **Expected-failure fixtures — do not expect these to pass.** Spec-correct fixtures that FAIL against at least one current implementation, parked in `EXPR/proposed/` (not discovered by the runner) so merging keeps the suite green. Per-fixture output, spec citation, and classification in `proposed/README.md`. Promote each fixture unchanged when its fix lands.

## Contents (5 fixtures, dual-implementation verified)

- **`expr2.1.1--int64-overflow-param-default.invalid`** — INT param `default: 2^63` accepted at `check` by BOTH implementations. The known live acceptance bug on the template-validation axis. Also blocked on restoring RFC 0005's dropped "64-bit Signed Integer Type" section to the published spec.
- **`expr2.1.1--int64-max-task-range.test`** — openjd-rs FALSELY REJECTS a valid task-range endpoint of 2^63−1 ("Integer overflow"); probing shows acceptance ends at 2^62−1. Python accepts and runs. The spec admits any int64. Rust-side bug, the mirror image of the acceptance bug.
- **`expr2.1.1--int64-overflow-task-range.invalid`** — task-range element 2^63: openjd-rs correctly rejects; the Python CLI accepts. Python-side bug.
- **`expr2.1.3--membership-element-type-mismatch.invalid`** — `"a" in [1, 2]` accepted by openjd-rs (evaluates to false at runtime); Python rejects at validation. §2.1.3 defines membership only as `(list[T], T)`. Rust-side bug.
- **`3.6.1--let-identifier-513.invalid`** — a 513-char let identifier accepted by BOTH implementations; schema §3.6.1 caps `<UserIdentifier>` at 512. The 512 accept-twin is green in the companion PR.

Companion to the green-fixture PR from branch `conformance-expr-lang-gaps`.

PR #165, opened 2026-08-02
test: Promote the center and isdigit function-library fixtures

leongdl:
Promotes the two remaining function-library fixtures out of `EXPR/jobs/proposed/` and into `jobs/`, where the runner discovers them. Both spec decisions they were waiting on have landed, and both fixtures pass.

## `expr2.2.4--center-odd-padding`

Section 2.2.4 does not say which side of an odd pad takes the extra space. Per @mwiebe's review: #177 settles ambiguities of this kind by matching Python. CPython's `'hi'.center(7)` is `'   hi  '` (extra space left), which is what the fixture already asserted as a strawman and is now the rule.

## `expr2.2.4--isdigit-unicode`

The classification functions carry Python `str` semantics, and openjd-rs#341 implemented that from CPython-generated tables. Per @mwiebe's review, the Unicode expectation now holds, so the fixture goes live asserting `isdigit` is true for U+0663 and U+0969.

Coverage note: #177 adds `expr2.2.4--string-classification-unicode`, which is a superset for U+0663 (same `ISDIGIT_ARABIC` marker). This fixture also covers U+0969, so it is not a pure duplicate, but it is reasonable to fold it into #177's fixture instead if you would rather have one file. Say the word and I will drop it.

## Fixture changes made while promoting

- Rewrote both headers: dropped the `PARKED` framing and the 2026-08 divergence tables, and stated the spec rule, the inputs, and what a wrong implementation prints.
- Wrote the digits as `\u0663` and `\u0969` escapes so the file stays ASCII. The runner reads fixtures with `open(path)` at the locale encoding, which mojibakes a UTF-8 literal on Windows. Same reason #177 uses escapes.
- Bracketed the `isdigit` assertions, matching the sibling `expr2.2.4--string-classification` fixture. The runner matches output as a substring, so an unbracketed value is a weaker assertion.
- Deleted `proposed/README-func-lib.md`. Nothing is left parked in it: the `repr_py` fixture was promoted on #162.

The third fixture from the original description, `expr2.2.6--repr-py-newline-roundtrip`, was already promoted on #162 and dropped from this branch.

### DIFF
--- a/conformance-tests/2023-09/EXPR/jobs/proposed/expr2.2.4--center-odd-padding.test.yaml
+++ b/conformance-tests/2023-09/EXPR/jobs/proposed/expr2.2.4--center-odd-padding.test.yaml
@@ -1,0 +1,36 @@
+# PARKED: spec-silent behaviour with a live implementation divergence.
+#
+# Expression Language §2.2.4 specifies center(s, width) only as "Center, pad
+# with spaces to width" — it does not say which side receives the extra space
+# when the padding is odd.
+#
+# Observed for center("hi", 7) (5 pad spaces to split), 2026-08 measurements:
+#   CPython str.center:   "   hi  "  (3 left, 2 right — extra space LEFT;
+#                          left = (width - len) // 2 + ((width - len) & width & 1))
+#   openjd-rs:            "   hi  "  (matches CPython since upstream #305 +
+#                          the RFC 0005 coercion sync; verified again on the
+#                          current upstream/main build)
+#   Python openjd CLI:    "  hi   "  (2 left, 3 right — now the OUTLIER)
+#
+# The expected: block below asserts the CPython/openjd-rs behaviour, which
+# has become the de facto answer. Do NOT move this fixture into jobs/ until
+# the spec picks a side ("follow CPython" would make this promotable and the
+# Python CLI's split a plain bug).
+template:
+  specificationVersion: jobtemplate-2023-09
+  extensions:
+  - EXPR
+  name: TestJob
+  steps:
+  - name: Step1
+    script:
+      actions:
+        onRun:
+          command: python
+          args:
+          - -c
+          - |
+            print(r'CENTER:[{{ center("hi", 7) }}]')
+expected:
+  output:
+  - "CENTER:[   hi  ]"
--- a/conformance-tests/2023-09/EXPR/jobs/proposed/expr2.2.4--isdigit-unicode.test.yaml
+++ b/conformance-tests/2023-09/EXPR/jobs/proposed/expr2.2.4--isdigit-unicode.test.yaml
@@ -1,0 +1,40 @@
+# PARKED: spec ambiguity — NO divergence between the openjd implementations.
+#
+# Expression Language §2.2.4 specifies isdigit(s) as "True if all characters
+# are digits and string is non-empty" without defining "digit": ASCII 0-9 or
+# Unicode digit category (Nd)? §2.2.5 explicitly gives \d Unicode semantics,
+# which pulls one way; both openjd implementations pull the other.
+#
+# Observed for isdigit("٣") (U+0663 ARABIC-INDIC DIGIT THREE):
+#   openjd-rs:              false  (ASCII-only digit test)
+#   Python openjd CLI:      false  (agrees — ASCII-only)
+#   CPython str.isdigit:    True   (Unicode Nd/No digit test — the host
+#                                   language differs from both)
+#
+# The expected: block below asserts the Unicode reading as a strawman,
+# matching §2.2.5's Unicode posture — i.e. the OPPOSITE of the current
+# de facto agreement. If the spec ratifies ASCII-only (the likely outcome
+# given both implementations agree), flip the expectation instead of
+# promoting as-is. Do NOT move this fixture into jobs/ until the spec
+# defines "digit" (and likewise "alphabetic"/"alphanumeric"/"whitespace"
+# for the sibling predicates).
+template:
+  specificationVersion: jobtemplate-2023-09
+  extensions:
+  - EXPR
+  name: TestJob
+  steps:
+  - name: Step1
+    script:
+      actions:
+        onRun:
+          command: python
+          args:
+          - -c
+          - |
+            print(r'ISDIGIT_ARABIC:{{ isdigit("٣") }}')
+            print(r'ISDIGIT_DEVANAGARI:{{ isdigit("३") }}')
+expected:
+  output:
+  - ISDIGIT_ARABIC:true
+  - ISDIGIT_DEVANAGARI:true
--- a/conformance-tests/2023-09/EXPR/jobs/proposed/expr2.2.6--repr-py-newline-roundtrip.test.yaml
+++ b/conformance-tests/2023-09/EXPR/jobs/proposed/expr2.2.6--repr-py-newline-roundtrip.test.yaml
@@ -1,0 +1,45 @@
+# PARKED: implementation bug found by the repr_py round-trip battery.
+#
+# Expression Language §2.2.6: repr_py "follows the behavior of Python's repr".
+# Python's repr("a\nb") is 'a\nb' (backslash-n ESCAPED, 7 characters, one line).
+# BOTH implementations (openjd-rs and the Python CLI — 2026-08-12 sweep) emit
+# a RAW newline inside the single quotes — which is not a valid
+# Python string literal at all:
+#
+#   ast.literal_eval on the emitted output raises
+#   "SyntaxError: unterminated string literal (detected at line 1)"
+#
+# So the output cannot be embedded in a generated Python script, which is the
+# entire purpose of repr_py. Backslash, quote, and tab handling round-trip
+# correctly (tab is emitted raw rather than as \t, which deviates from repr()
+# but still parses back byte-identical); newline is the case that produces
+# invalid output. Classification: implementation bug in BOTH implementations
+# (spec is
+# explicit via the repr() reference). This fixture is the regression test to
+# move into jobs/ once fixed.
+template:
+  specificationVersion: jobtemplate-2023-09
+  extensions:
+  - EXPR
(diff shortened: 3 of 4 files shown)


conformance-tests/2023-09/EXPR/proposed/expr2.2.4--center-odd-padding.test.yaml
leongdl, line 1: **Quorum verdict: NOT-GOOD (4/5 — three independent NOT-GOODs converged on the README)** — spec: EL §2.2.4 center() is silent on the odd-split side, so 'spec question' is the right classification. But the README entry is wrong on the facts: (1) it claims both implementations return right-heavy '  hi   ' with 'no divergence' — the 2026-08-12 sweep shows openjd-rs changed (upstream #305 + RFC 0005 coercion sync) and now emits CPython's '   hi  '; only the Python CLI is right-heavy now (fails PYTHON only); (2) it misstates CPython — verified: 'hi'.center(7) == '   hi  ' (extra space LEFT; left = pad//2 + (pad & width & 1) is not Python's formula); (3) it declares 'the fixture's left-heavy expectation was wrong' while the committed fixture still asserts '   hi  ' — which now matches both CPython and openjd-rs. Action: rewrite the README entry to the sweep facts; the fixture's expectation can stand as the de facto pin, with the spec decision (document the CPython split) as the promotion gate.

conformance-tests/2023-09/EXPR/proposed/expr2.2.4--isdigit-unicode.test.yaml
leongdl, line 1: **Quorum verdict: GOOD-WITH-NITS** — spec: EL §2.2.4 leaves 'digit' undefined while §2.2.5 gives \d explicit Unicode semantics — correctly classified as a spec ambiguity, and the README's dual-run (both implementations ASCII-only, agreeing) matches the sweep. Nits: (1) the fixture header claims 'live implementation divergence' — there is none; both agree, the divergence is vs host-language str.isdigit — align the header with the README; (2) the strawman asserts Unicode-true, the OPPOSITE of the de facto agreement — if the spec ratifies ASCII (likely), the expectation flips; two reviewers suggest either flipping now or leaving this out of the inventory entirely since it prejudges an open spec decision rather than pinning a bug.

conformance-tests/2023-09/EXPR/jobs/proposed/expr2.2.6--repr-py-newline-roundtrip.test.yaml
leongdl, line 1: **Quorum verdict: GOOD (5/5, one stale comment)** — spec: EL §2.2.6 (repr_py follows Python repr; the spec's own example escapes \n) — genuinely mandated, and the README's 'bug in BOTH implementations, dual-run verified' matches the sweep. The ast.literal_eval round-trip asserts semantics rather than byte-exact escaping — well built for promotion. Fix: the fixture's inline comment still says 'bug in openjd-rs' — update to both.

leongdl:
Quorum review (5 independent agents). 3 fixtures + README. Net: 1 GOOD, 1 GOOD-WITH-NITS, 1 NOT-GOOD requiring a README rewrite before merge. The center-odd-padding README is internally contradictory and factually wrong about CPython (verified by execution: 'hi'.center(7) == '   hi  ', extra space LEFT); per the 2026-08-12 sweep, openjd-rs now matches CPython and the PYTHON CLI is the outlier — the fixture's committed expectation actually matches the de facto answer, and the README is the defect. Cross-PR blocker: this branch and PR #164 both create EXPR/proposed/README.md (verified add/add conflict).

leongdl:

- **`expr2.2.4--center-odd-padding` README/header rewritten on the measured facts** (the NOT-GOOD): CPython's `'hi'.center(7)` is `'   hi  '` — extra space LEFT (verified by execution, three independent runs this review); current openjd-rs matches CPython (re-verified against an upstream/main build: emits `'   hi  '`); the Python openjd CLI is now the right-heavy outlier. The fixture's committed expectation was correct all along — the old README's "both implementations are right-heavy / the expectation was wrong" claims are retracted in the new text.
- `expr2.2.4--isdigit-unicode`: header reframed as spec ambiguity with **no divergence between the openjd implementations** (both ASCII-only; the divergence is vs host CPython), and the strawman's flip-on-ASCII-ratification path is documented.
- `expr2.2.6--repr-py-newline-roundtrip`: stale "bug in openjd-rs" comment corrected to BOTH implementations.
- Restructured to kind-level parking (`EXPR/jobs/proposed/`) with a family-named README (`README-func-lib.md`) — removes the add/add merge conflict with #164.

conformance-tests/2023-09/EXPR/jobs/proposed/expr2.2.4--isdigit-unicode.test.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - name: Step1
+    script:
+      actions:
+        onRun:
+          command: python
+          args:
+          - -c
+          - |
+            print(r'ISDIGIT_ARABIC:{{ isdigit("٣") }}')
+            print(r'ISDIGIT_DEVANAGARI:{{ isdigit("३") }}')
```
mwiebe, line 36: This should be working now with https://github.com/OpenJobDescription/openjd-rs/pull/341 merged

conformance-tests/2023-09/EXPR/jobs/proposed/expr2.2.4--center-odd-padding.test.yaml
```diff
@@ -0,0 +1,36 @@
...
+#                          left = (width - len) // 2 + ((width - len) & width & 1))
+#   openjd-rs:            "   hi  "  (matches CPython since upstream #305 +
+#                          the RFC 0005 coercion sync; verified again on the
+#                          current upstream/main build)
+#   Python openjd CLI:    "  hi   "  (2 left, 3 right — now the OUTLIER)
+#
+# The expected: block below asserts the CPython/openjd-rs behaviour, which
+# has become the de facto answer. Do NOT move this fixture into jobs/ until
+# the spec picks a side ("follow CPython" would make this promotable and the
+# Python CLI's split a plain bug).
```
mwiebe, line 18: In https://github.com/OpenJobDescription/openjd-specifications/pull/177 we went with matching Python whenever there was ambiguity about it, I think that's a good approach to follow here too.

conformance-tests/2023-09/EXPR/jobs/proposed/expr2.2.4--isdigit-unicode.test.yaml
leongdl: Nice!

conformance-tests/2023-09/EXPR/jobs/proposed/expr2.2.4--center-odd-padding.test.yaml
leongdl: Sounds good, will update the test.

conformance-tests/2023-09/EXPR/jobs/proposed/expr2.2.4--isdigit-unicode.test.yaml
leongdl: Correct, and promoted. Moved to `jobs/` in fb2898c with the header rewritten to state the Python-`str` rule rather than the old ASCII/Unicode open question, crediting openjd-rs#341.

Two changes while promoting: the digits are now `\u0663` and `\u0969` escapes so the file stays ASCII (the runner opens fixtures at the locale encoding, which mojibakes a UTF-8 literal on Windows, which is the same reason #177 uses escapes), and the assertions are bracketed to match `expr2.2.4--string-classification`, since the runner matches output as a substring.

One thing worth your call: #177's `expr2.2.4--string-classification-unicode` already asserts `ISDIGIT_ARABIC:[true]` for U+0663 under the same marker name. This fixture adds U+0969, so it is not a pure duplicate, but if you would rather carry one file I am happy to drop it and let #177 own the coverage.

conformance-tests/2023-09/EXPR/jobs/proposed/expr2.2.4--center-odd-padding.test.yaml
leongdl: Applied. Promoted to `jobs/` in fb2898c on the match-Python rule from #177.

The committed expectation already was CPython's, so the assertion is unchanged at `CENTER:[   hi  ]` (3 left, 2 right). What changed is the framing: the header no longer parks the fixture behind an open spec question, it cites 2.2.4's silence, names #177's match-Python resolution, and says what a right-heavy implementation prints instead. The 2026-08 divergence table is gone from the header, so nothing in the fixture now claims anything about the Python CLI's split.

leongdl:
CI on fb2898c: both promoted fixtures are now discovered by the runner and pass on every lane that has completed.

| Lane | OS | Result | Suite total |
|---|---|---|---|
| Conformance Tests (Rust) | ubuntu | `✓ expr2.2.4--center-odd-padding`, `✓ expr2.2.4--isdigit-unicode` | 1184 passed, 0 failed |
| Conformance Tests (Rust) | windows | both `✓` | 1139 passed, 0 failed |
| Conformance Tests (Python) | ubuntu | both `✓` | 1184 passed, 0 failed |
| Conformance Tests (Python) | windows | both `✓` | 1139 passed, 0 failed |

macOS (Rust lane) has been queued for 45 minutes without starting, which is a hosted-runner backlog rather than a result. Both fixtures are platform-independent string operations and pass on the other four lanes.

Worth recording because it retracts an earlier measurement: the old `center-odd-padding` header claimed the Python openjd CLI was the right-heavy outlier and would fail this fixture. It passes. Whatever the 2026-08 sweep measured, the current Python CLI agrees with CPython and openjd-rs on `'hi'.center(7) == '   hi  '`, so there is no live divergence left on either fixture.

mwiebe (approved):

PR #158, opened 2026-08-02
test: Add 13 extended-parameter-type conformance fixtures

leongdl:
## Summary

13 extended-parameter-type fixtures from a coverage audit of RFC 0007. Suite: **EXPR 350 → 363 passing, 0 failing**; base unchanged.

## Areas covered

- **int64 accept twins on list element types**: `LIST[INT]` element and `LIST[LIST[INT]]` inner element at 2^63−1 accepted — the data-axis boundary the audit found untested for every container type.
- **First-ever `LIST[BOOL]` runtime coverage** (the audit found `EXPR/jobs/` had no `2.15--*` file at all): element values reaching the task, and interpolation rendering. Note: these two fixtures pass on openjd-rs and currently FAIL on the Python CLI (`TypeError: List contains incompatible types: bool, string`) — they are live divergence catchers, green against the reference implementation.
- **Mixed-case type spellings**: `LiSt[iNt]` and friends accepted with EXPR (job parameters).
- **Normative prohibitions**: `BOOL` + `allowedValues` rejected; `item:` on `LIST[BOOL]` rejected; `LIST[LIST[LIST[INT]]]` triple nesting rejected — all stated in the schema, none previously tested.
- **`LIST[PATH]` element constraints**: `item.allowedValues` accept + violation pair (previously appeared in zero fixtures).
- **Indexed-element property access**: `{{Param.Paths[0].stem}}` — the subscript route of the known property-access bug class; verified NOT affected.
- **`RANGE_EXPR` endpoint overflow reject** and a **`LIST[STRING]` wrong-item-type negative** (numeric elements rejected, no silent YAML coercion — verified on both implementations).

## Companion PR

Five spec-correct fixtures that FAIL today (list-element 2^63 acceptance in both implementations, the openjd-rs 2^62 endpoint cap, task-parameter case-insensitivity unimplemented in both, base lowercase type acceptance in openjd-rs) are in the expected-failures PR (branch `conformance-param-types-expected-failures`).

PR #166, opened 2026-08-02
test: Add a §2 task parameter type case-insensitivity fixture

leongdl:
Adds one conformance fixture: `2023-09/EXPR/job_templates/3.4.1--task-param-type-case-insensitive.yaml`.

### What it pins

Template Schemas §2 makes job **and task** parameter type names case-insensitive when the EXPR extension is enabled. The case was implemented for job parameter types only, so a task parameter spelled `type: int` was rejected with EXPR declared:

```
steps[0] -> parameterSpace -> taskParameterDefinitions[0]:
	Input tag 'int' found using 'type' does not match any of the expected tags: INT, FLOAT, STRING, PATH, CHUNK[INT]
```

The fixture covers the four task parameter types of §3.4.1 — `int`, `Float`, `sTrInG`, `pAtH` — and must validate.

### Status: red until two releases ship

Both implementations have merged the fix, and neither has released it:

| Lane | Installs | Contains the fix? |
|---|---|---|
| Python | `pip install openjd-cli` → openjd-model 0.11.9 | No. [openjd-model-for-python#350](https://github.com/OpenJobDescription/openjd-model-for-python/pull/350) merged 2026-09-09, after the 0.11.9 tag |
| Rust | `cargo install openjd-cli` → 0.1.14 | No. [openjd-rs#364](https://github.com/OpenJobDescription/openjd-rs/pull/364) merged, and 0.1.14 predates it |

Measured, not inferred: the released Python CLI rejects the fixture, and a build of openjd-rs `main` accepts it.

Neither release is queued, and each needs a deliberate act:

- openjd-model-for-python: dispatch `release_bump.yml` (manual `workflow_dispatch`), which lands a CHANGELOG bump on mainline and triggers publish.
- openjd-rs: merge the standing release PR [#365](https://github.com/OpenJobDescription/openjd-rs/pull/365), which does contain #364.

Once both are on PyPI and crates.io, this goes green with no further change here.

### Changes from the previous revision

This branch previously parked four fixtures under `job_templates/proposed/`, which the runner does not scan — collection uses a non-recursive `directory.glob("*.yaml")` per kind directory, so nothing in `proposed/` ever ran.

The case-insensitivity fixture is now an ordinary conformance test, and `proposed/` plus its README are gone. The branch is also rebased onto mainline, which clears an unrelated failure: it carried the pre-rename `3.5--env-script-onexit-only.invalid.yaml`, corrected upstream in 563c6ec.

The other three parked fixtures were dropped rather than promoted, because each still fails:

| Fixture | Expected | openjd-rs `main` | openjd-model 0.11.9 |
|---|---|---|---|
| `2.10--range-expr-endpoint-int64-max` | valid | rejects | accepts |
| `2.13--list-int-item-int64-overflow.invalid` | reject | accepts | accepts |
| `2.16--list-list-int-inner-item-int64-overflow.invalid` | reject | accepts | accepts |

They pin genuine int64-bound defects — bounds are enforced for expression literals but not for values arriving as YAML data — and the two `.invalid` ones additionally depend on restoring the overflow-is-an-error text that was dropped from the published spec. Better raised on their own once those are addressed than carried in a PR that is otherwise ready.

### Not covered, and worth adding

Nothing in the suite pins the other half of §2: that a lowercase task parameter type is still **rejected** without EXPR. Both implementations gate the fold on the extension, so a regression dropping that gate would pass the whole suite.

### DIFF
--- a/conformance-tests/2023-09/EXPR/job_templates/proposed/2.10--range-expr-endpoint-int64-max.yaml
+++ b/conformance-tests/2023-09/EXPR/job_templates/proposed/2.10--range-expr-endpoint-int64-max.yaml
@@ -1,0 +1,19 @@
+# Section 2.10 / 3.4.1.1.1: RANGE_EXPR endpoints at the int64 boundary.
+# 9223372036854775807 is 2^63-1, the largest representable endpoint.
+specificationVersion: jobtemplate-2023-09
+extensions:
+- EXPR
+name: TestJob
+parameterDefinitions:
+- name: Frames
+  type: RANGE_EXPR
+  default: "9223372036854775806-9223372036854775807"
+steps:
+- name: Step1
+  script:
+    actions:
+      onRun:
+        command: python
+        args:
+        - "-c"
+        - "print()"
--- a/conformance-tests/2023-09/EXPR/job_templates/proposed/2.13--list-int-item-int64-overflow.invalid.yaml
+++ b/conformance-tests/2023-09/EXPR/job_templates/proposed/2.13--list-int-item-int64-overflow.invalid.yaml
@@ -1,0 +1,23 @@
+# PARKED (see README-param-types.md): a LIST[INT] default element of
+# 9223372036854775808 (2^63) exceeds the int64 range (Expression Language
+# 1.2.1 — the bound comes from the EXPR type table, not section 2.13
+# itself) and must be rejected. Both implementations currently accept it.
+# Accept twin at 2^63-1: 2.13--list-int-item-int64-max.yaml
+# (param-types-gaps PR).
+specificationVersion: jobtemplate-2023-09
+extensions:
+- EXPR
+name: TestJob
+parameterDefinitions:
+- name: Values
+  type: LIST[INT]
+  default: [9223372036854775808]
+steps:
+- name: Step1
+  script:
+    actions:
+      onRun:
+        command: python
+        args:
+        - "-c"
+        - "print()"
--- a/conformance-tests/2023-09/EXPR/job_templates/proposed/2.16--list-list-int-inner-item-int64-overflow.invalid.yaml
+++ b/conformance-tests/2023-09/EXPR/job_templates/proposed/2.16--list-list-int-inner-item-int64-overflow.invalid.yaml
@@ -1,0 +1,23 @@
+# PARKED (see README-param-types.md): a LIST[LIST[INT]] inner element of
+# 9223372036854775808 (2^63) exceeds the int64 range (Expression Language
+# 1.2.1) — the same int64-in-data defect as 2.13, one container level
+# deeper, guarding nested validation paths. Both implementations currently
+# accept it. Accept twin: 2.16--list-list-int-inner-item-int64-max.yaml
+# (param-types-gaps PR).
+specificationVersion: jobtemplate-2023-09
+extensions:
+- EXPR
+name: TestJob
+parameterDefinitions:
+- name: Matrix
+  type: LIST[LIST[INT]]
+  default: [[9223372036854775808]]
+steps:
+- name: Step1
+  script:
+    actions:
+      onRun:
+        command: python
+        args:
+        - "-c"
+        - "print()"
--- a/conformance-tests/2023-09/EXPR/job_templates/proposed/3.4.1--task-param-type-case-insensitive.yaml
+++ b/conformance-tests/2023-09/EXPR/job_templates/proposed/3.4.1--task-param-type-case-insensitive.yaml
@@ -1,0 +1,31 @@
+# With EXPR enabled, task parameter type names are case-insensitive
+# (Template Schemas section 2: "job parameter and task parameter type names
+# become case-insensitive"). Task parameter types are INT, FLOAT, STRING,
+# and PATH (section 3.4.1); LIST forms are not task parameter types.
+specificationVersion: jobtemplate-2023-09
+extensions:
+- EXPR
+name: TestJob
+steps:
+- name: Step1
+  parameterSpace:
+    taskParameterDefinitions:
+    - name: Frame
+      type: int
+      range: "1-3"
+    - name: Scale
+      type: Float
+      range: ["1.0", "2.0"]
+    - name: Layer
+      type: sTrInG
+      range: ["fg", "bg"]
+    - name: Scene
+      type: pAtH
+      range: ["/tmp/a.blend", "/tmp/b.blend"]
+  script:
+    actions:
+      onRun:
+        command: python
+        args:
+        - "-c"
+        - "print()"
--- a/conformance-tests/2023-09/EXPR/job_templates/proposed/README-param-types.md
+++ b/conformance-tests/2023-09/EXPR/job_templates/proposed/README-param-types.md
@@ -1,0 +1,87 @@
+# Proposed fixtures (extended parameter types) — job_templates
+
+These fixtures are believed spec-correct but fail against at least one
+current reference implementation. Placement is kind-level
+(`<component>/<kind>/proposed/`): the runner does not scan `proposed/`, and
+promotion is a move up one directory unchanged. This README is named per
+fixture family (`README-param-types.md`) because other expected-failures
+PRs park fixtures in this same directory with their own READMEs.
+
+Observations are per-implementation (`rs` = openjd-rs, `py` = the Python
+openjd CLI); dual results from the 2026-08-12 sweep, rs re-verified against
+the current upstream/main build.
+
+## 2.13--list-int-item-int64-overflow.invalid.yaml
+## 2.16--list-list-int-inner-item-int64-overflow.invalid.yaml
+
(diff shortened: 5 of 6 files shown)


conformance-tests/2023-09/EXPR/job_templates/proposed/2.10--range-expr-endpoint-int64-max.yaml
leongdl, line 1: **Quorum verdict: GOOD (5/5)** — spec: §2.10 + §3.4.1.1.1 + EL §1.2.1 (2^63-1 endpoints must validate). Pins openjd-rs's undocumented 2^62 endpoint cap (rs-only per sweep); fix widens acceptance only. The README's note that the companion negative (PR #158's 2.10--range-expr-endpoint-int64-overflow) passes for the wrong reason until this accept is green is exactly right — promote as a pair. Nit: the quoted validation error looks pydantic-style (Python) for an rs-attributed bug — double-check which CLI produced it, and state that Python passes.

conformance-tests/2023-09/EXPR/job_templates/proposed/2.13--list-int-item-int64-overflow.invalid.yaml
leongdl, line 1: **Quorum verdict: GOOD-WITH-NITS** — spec: §2.13 + EL §1.2.1 (LIST[INT] element bound). Matches sweep (fails BOTH — say so; README says 'the current openjd CLI' without attribution). Nits: only fixture family in the five PRs without in-file header comments — add one (the bound comes from the EL type table, not §2.13); carry the same 'pending restoration of the RFC 0005 overflow text' hedge the expr-lang README uses for the identical bound. Accept twin: PR #158's 2.13--list-int-item-int64-max.

conformance-tests/2023-09/EXPR/job_templates/proposed/2.16--list-list-int-inner-item-int64-overflow.invalid.yaml
leongdl, line 1: **Quorum verdict: GOOD-WITH-NITS** — spec: §2.16 (inner element, one container deeper — a declared sweep axis, kept despite likely sharing 2.13's recursive code path). Same nits as 2.13: no header comment, no per-implementation attribution, needs the spec-restoration hedge. Accept twin: PR #158's 2.16--list-list-int-inner-item-int64-max.

conformance-tests/2023-09/EXPR/job_templates/proposed/3.4.1--task-param-type-case-insensitive.yaml
leongdl, line 1: **Quorum verdict: GOOD (5/5, one attribution fix)** — spec: §2 intro (with EXPR, job AND task parameter type names are case-insensitive — the sentence explicitly covers task types). Acceptance-widening only; queue environments unaffected (they carry job params). Promotion-ready. Fix: the README quotes a serde/Rust-style error and diagnoses rs-specifically, but the sweep shows BOTH implementations fail — add Python's observation. Inventory note: the reject twin (task-param lowercase WITHOUT EXPR) doesn't exist anywhere — consider adding it.

conformance-tests/2023-09/base/job_templates/proposed/2--type-lowercase-string.invalid.yaml
leongdl, line 1: **Quorum verdict: GOOD-WITH-NITS (dedupe action)** — spec: §2 (base type names case-sensitive; the gate is otherwise undetectable suite-wide — good justification). Matches sweep (rs-only; Python correctly rejects — README omits this, its PR #163 sibling states it). Action: duplicates PR #163's base/proposed/2--type-lowercase.invalid.yaml — same construct, same defect, two different proposed/ paths; if both PRs land the same pin exists twice. Keep one or differentiate the type token.

leongdl:
Quorum review (5 independent agents). 5 fixtures + 2 READMEs. Net: 3 GOOD, 2 GOOD-WITH-NITS. This branch's kind-level placement (EXPR/job_templates/proposed/, base/job_templates/proposed/) is the better convention of the two in use — promotion is 'move up one directory unchanged' — and the quorum recommends the other four expected-failures branches adopt it. Actions: dedupe 2--type-lowercase-string against PR #163's identical fixture; add per-implementation attribution to the README (sweep shows several entries fail BOTH or rs-only where the README is vague); add the missing header comments to the two list-overflow fixtures and carry the same spec-restoration hedge the expr-lang README uses for the identical int64 bound.

leongdl:

- **Dropped** `2--type-lowercase-string.invalid.yaml` and its README — it duplicated #163's `2--type-lowercase.invalid.yaml` (same construct, same rs-only defect); that branch's copy is the single source.
- README renamed to `README-param-types.md` (family-named, kind-level convention) so co-located expected-failures PRs can't add/add-conflict on a shared README path.
- Per-implementation attribution added: `2.13`/`2.16` list-overflow and `3.4.1` task-param case-insensitivity fail **both** implementations; `2.10--range-expr-endpoint-int64-max` is **rs-only** (the quoted "Model validation error" format is genuinely openjd-rs output — same format as pydantic — re-verified against the current upstream/main build, where all four defects still reproduce).
- Added the dropped-RFC-0005-overflow-text hedge to the 2.13/2.16 entries (matching the expr-lang README) and header comments to both list-overflow fixtures.
- Noted the missing reject twin (task-param lowercase type WITHOUT EXPR) as a future addition.

conformance-tests/2023-09/EXPR/job_templates/proposed/2.10--range-expr-endpoint-int64-max.yaml
```diff
@@ -0,0 +1,19 @@
...
+  default: "9223372036854775806-9223372036854775807"
+steps:
+- name: Step1
+  script:
+    actions:
+      onRun:
+        command: python
+        args:
+        - "-c"
+        - "print()"
```
leongdl: Lets print this and assert the output.

leongdl: Do we have a negative - case as well, try to group them into one test set.

conformance-tests/2023-09/EXPR/job_templates/proposed/2.16--list-list-int-inner-item-int64-overflow.invalid.yaml
```diff
@@ -0,0 +1,23 @@
...
+# accept it. Accept twin: 2.16--list-list-int-inner-item-int64-max.yaml
+# (param-types-gaps PR).
+specificationVersion: jobtemplate-2023-09
+extensions:
+- EXPR
+name: TestJob
+parameterDefinitions:
+- name: Matrix
+  type: LIST[LIST[INT]]
+  default: [[9223372036854775808]]
```
leongdl: Also add a negative cases for coverage.

conformance-tests/2023-09/EXPR/job_templates/proposed/README-param-types.md
```diff
@@ -0,0 +1,87 @@
+# Proposed fixtures (extended parameter types) — job_templates
```
leongdl: Delete these files.

conformance-tests/2023-09/EXPR/job_templates/proposed/2.10--range-expr-endpoint-int64-max.yaml
```diff
@@ -0,0 +1,19 @@
+# Section 2.10 / 3.4.1.1.1: RANGE_EXPR endpoints at the int64 boundary.
```
leongdl: The four steps
1. Merge PR 350 — the Python code fix. Two blockers right now: it's BEHIND mainline so it needs updating, and REVIEW_REQUIRED with no approval. Checks pass.

2. Release openjd-model to PyPI. This is the step people forget. The specs Python lane runs pip install openjd-cli, so it tests whatever is published. Release fires on a push to mainline touching [CHANGELOG.md](http://changelog.md/), produced by release_bump.yml which is workflow_dispatch only — a human has to trigger it. Until then the lane still resolves openjd-model 0.11.9, which has the bug.

3. Release openjd-rs. The Rust lane runs cargo install openjd-cli, and #364 is merged but unreleased. The good news: release PR #365 already contains it — I verified 361b2c5 is an ancestor of that branch's 40c29ad. Merging #365 publishes the fix.

4. Merge PR 166 and promote the fixture. Merging 166 lands the file on mainline still inside proposed/, where the non-recursive glob can't see it. PR 166's own README states the rule: "promotion is a move up one directory unchanged". So promotion is a separate one-line move from EXPR/job_templates/proposed/ to EXPR/job_templates/.

seant-aws (approved):

mwiebe (approved):

PR #167, opened 2026-08-06
fix(rfc): make RFC 0008's Docker example a valid template

seant-aws:
## PR for other purpose

RFC 0008's Basic Example is not a valid environment template. It interpolates
`WrappedAction.*` inside the wrap script's `embeddedFiles` `data:`, the shape
`conformance-tests/2023-09/WRAP_ACTIONS/env_templates/4--wrappedaction-in-embedded-file.invalid.yaml`
declares invalid — embedded files are shared by every action in the script,
including the environment's own `onEnter`/`onExit`, where the wrap variables do
not exist. `openjd-model` rejects the example as published:

```
environment -> script -> embeddedFiles[1] -> data:
    Variable WrappedAction.Environment does not exist at this location.
    Variable WrappedAction.Command does not exist at this location.
    Variable WrappedAction.Args does not exist at this location.
```

`wrap.sh` becomes static and receives the wrapped action as argv, in the layout
`[-e KEY=value ...] -- <command> [args ...]`, with every `WrappedAction.*`
reference moved into the hooks' `args:` where the scope rule permits it. Because
the values arrive as argv rather than as text a shell must parse, no quoting is
required on this path.

Two further defects in the same example, both found by running it:

- `docker container exec` placed the `-e` options **after** the container
  argument; docker parses everything after the container argument as the command
  to run.
- `docker container stop --timeout` is not a valid flag — the option is
  `-t`/`--time` — so `onExit` exited 125 and leaked the container.

Prose is updated to match the example. Argv forwarding is now presented as the
recommended default, with `repr_sh` scoped to hooks whose target takes a shell
command line (`ssh host '<line>'`, `sh -c`, batch scheduler submit strings).
Security Considerations no longer suggests reconstructing the command line
"inside an embedded shell script", which is the pattern the scope rule forbids,
and its cross-reference to the Basic Example's `repr_sh` forms is corrected.
`repr_sh` keeps its section and its rationale — it moves from "always" to "when
the target parses a shell command line".

### How this was verified

The example was extracted from this document, validated as both an
`EnvironmentTemplate` and a `JobTemplate` with `EXPR` and `WRAP_ACTIONS`
enabled, and then executed:

- **as published** → rejected, with the three validation errors above
- **after this change** → accepted, and runs: all three wrap hooks fire, the
  task executes inside the container (`uid=0`, `/.dockerenv` present), shell
  metacharacters survive intact (`[a b c] [quote'q] [meta;|&>x]`), a wrapped
  command with an empty `args` list contributes zero argv entries, and the
  container is torn down with no leak.

Tested against `openjd-model` 0.11.2, `openjd-sessions` 0.10.11, and docker
25.0.14. No conformance fixtures or samples are modified, so no CI job is
affected.

### Possible follow-up

The scope rule is enforced for `EnvironmentTemplate` and `JobTemplate`, but not
for the bare `Environment` model. A consumer that parses a single environment on
its own — as a scheduler handing one environment to a session does — accepts the
invalid shape and fails at run time instead, with an error naming
`WrappedAction.Command`, which reads as a wrap-actions bug rather than a template
one. Happy to add a conformance fixture for that entry point as a separate PR if
that would be useful.

---

### DIFF
--- a/rfcs/0008-environment-wrap-actions.md
+++ b/rfcs/0008-environment-wrap-actions.md
@@ -52,15 +52,32 @@ environment:
       onEnter:
         command: bash
         args: ["{{Env.File.Enter}}"]
+      # Every WrappedAction.* reference lives here, in the hooks' args, where
+      # the wrap variables are in scope. wrap.sh itself stays static.
       onWrapEnvEnter:
         command: bash
-        args: ["{{Env.File.Wrap}}"]
+        args:
+          - "{{Env.File.Wrap}}"
+          - "{{ flatten([['-e', e] for e in WrappedAction.Environment]) }}"
+          - "--"
+          - "{{WrappedAction.Command}}"
+          - "{{WrappedAction.Args}}"
       onWrapTaskRun:
         command: bash
-        args: ["{{Env.File.Wrap}}"]
+        args:
+          - "{{Env.File.Wrap}}"
+          - "{{ flatten([['-e', e] for e in WrappedAction.Environment]) }}"
+          - "--"
+          - "{{WrappedAction.Command}}"
+          - "{{WrappedAction.Args}}"
       onWrapEnvExit:
         command: bash
-        args: ["{{Env.File.Wrap}}"]
+        args:
+          - "{{Env.File.Wrap}}"
+          - "{{ flatten([['-e', e] for e in WrappedAction.Environment]) }}"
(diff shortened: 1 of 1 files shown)


leongdl (approved):
Thanks! Good catches and cleanup

crowecawcaw (approved):

PR #172, opened 2026-08-14
fix: do not cap <IntRangeExpr> expansion at the list-form limit

leongdl:
## What

Two conformance tests apply §3.4.1.1's 1024-element list cap to an `<IntRangeExpr>`, which the
specification does not constrain by element count. This corrects them to test the cap on the form it
is actually defined for, adds a case covering a wide expression, and states the rule explicitly in
§3.4.1.1.1 so the ambiguity does not recur.

## Why

§3.4.1.1 item 4 places the 1024-element maximum on `<IntRangeList>`, and §3.4.1.2 / §3.4.1.3 do the
same for `<FloatRangeList>` / `<StringRangeList>`. §3.4.1.1.1 `<IntRangeExpr>` states one constraint
only — that no two ranges may overlap — and its stated motivating use-case is "providing a succinct
way to describe a frame range". A 1024-value ceiling contradicts that use-case: `1-5000` is an
ordinary frame range.

The reference implementation agrees. `openjd-model-for-python` validated the expression form with the
grammar alone, and applied `max_length=1024` only to `IntRangeList`, `FloatRangeList` and
`StringRangeList`, through 0.11.0 — every release up to July 2026. The element cap first reached the
expression form in 0.11.1, and the Python repository does not run this suite in CI, so the conflict
went unnoticed on that side.

There is also a practical cost to capping the expression form: it silently caps the number of Tasks a
Step can define through a single parameter, pre-empting the Task-count limits a host service sets and
may raise per customer. A host that wants to bound fan-out should bound the Task count directly.

## Changes

Conformance tests, `2023-09/base/job_templates/`:

- `3.4--too-many-range-items.invalid.yaml` — was `range: "1-1025"` (an expression). Now an
  `<IntRangeList>` of 1025 items, so it tests the cap that §3.4.1.1 item 4 defines. This matches the
  existing sibling tests `3.4.1--float-too-many-range-items.invalid.yaml`,
  `3.4.1--path-too-many-range-items.invalid.yaml` and
  `3.4.1--string-too-many-range-items.invalid.yaml`, all of which already use list form. INT was the
  only type whose list-form cap was untested.
- `3.4--max-range-items.yaml` — was `range: "1-1024"` (an expression). Now an `<IntRangeList>` of
  exactly 1024 items, giving the valid boundary counterpart in the same form.
- `3.4--wide-int-range-expression.yaml` — new. `range: "1-5000"` must be valid.

Specification text:

- `wiki/2023-09-Template-Schemas.md` §3.4.1.1.1 — one paragraph stating that the number of values an
  `<IntRangeExpr>` expands to is unconstrained, that §3.4.1.1's maximum does not apply to this form,
  and that an implementation needing to bound Task count should limit the Task count rather than the
  expression.

## Verification

`2023-09/*` against a `openjd-rs` build carrying the matching implementation fix: **1160 passed, 0
failed**. Before this change the same build reported 1158 passed / 1 failed, the failure being
`3.4--too-many-range-items.invalid.yaml`.

The corrected invalid test still fails validation as intended, which confirms the list-form cap
remains enforced — only the expression form is unbounded.

## Related

- `openjd-model-for-python` #327 — restores the pre-0.11.1 behavior (merged)
- `openjd-rs` #318 — same change in Rust; blocked on this PR, since the suite currently asserts the
  opposite


crowecawcaw (approved):

seant-aws (approved):

godobyte:
~~Is the failing CI expected?~~

Synced offline, the CI needs respective fix to be merged to pass.

godobyte (approved):

PR #173, opened 2026-08-15
fix: correct upper-Z EXPR conformance fixture

### LINKED ISSUE #171: Conformance: re-backslash-lower-z and upper-Z EXPR fixtures are byte-identical, so backslash-Z is never tested
### Summary

In `conformance-tests/2023-09/EXPR/job_templates/`:

- `expr2.2.5--re-backslash-lower-z.invalid.yaml` and
  `expr2.2.5--re-backslash-upper-Z.invalid.yaml` are **byte-for-byte identical**
  (git blob `92d1159` for both on current `main`; md5 `4e4ed82bbf9426c8580795ca03bcaebb`).
- Both contain the pattern `r"llo\z"` — lowercase `z` — despite the "upper-Z" filename.
- Consequently nothing in the 2023-09 EXPR suite tests `\Z` at all, although RFC 0006
  excludes it from the dialect just as it excludes `\z`.

### Suggested fix

Change the pattern in the `upper-Z` fixture to `r"llo\Z"` so the two fixtures test what
their names say, and `\Z` rejection gains coverage.

Sanjays2402:
Closes #171

The upper-Z EXPR invalid fixture was byte-for-byte identical to the lower-z fixture and incorrectly exercised `\\z`. It now uses `\\Z`, so the conformance corpus tests the upper-case escape named by the file.

This change was prepared with AI assistance; the regression assertion was run locally and fails without the fix.

---

### DIFF
--- a/conformance-tests/2023-09/EXPR/job_templates/expr2.2.5--re-backslash-upper-Z.invalid.yaml
+++ b/conformance-tests/2023-09/EXPR/job_templates/expr2.2.5--re-backslash-upper-Z.invalid.yaml
@@ -10,4 +10,4 @@ steps:
         command: python
         args:
         - -c
-        - print(r'{{ re_search("hello", r"llo\z") }}')
+        - print(r'{{ re_search("hello", r"llo\Z") }}')


leongdl (changes requested):
Please do a `git commit --amend -s` to sign the commit.

conformance-tests/2023-09/EXPR/job_templates/expr2.2.5--re-backslash-upper-Z.invalid.yaml
```diff
@@ -10,4 +10,4 @@ steps:
         command: python
         args:
         - -c
-        - print(r'{{ re_search("hello", r"llo\z") }}')
+        - print(r'{{ re_search("hello", r"llo\\Z") }}')
```
leongdl, line 13: This should be a single backslash: `r"llo\Z"`. In an EXPR raw string, `\\Z` is two literal backslashes — as a regex that's an escaped backslash plus a literal `Z`, which is valid in the dialect, so the fixture no longer tests rejection of the `\Z` anchor. Compare the lower-z sibling (`r"llo\z"`) and the backreference fixture (`r"(a)\1"`), which both use single backslashes.

leongdl (changes requested):
Thanks for fixing this — the diagnosis is correct (both fixtures are the identical blob `92d1159` on mainline, so `\Z` currently has no coverage), but the replacement pattern has one backslash too many.

The file now contains `r"llo\\Z"` (two literal backslashes). Since EXPR `r"..."` is a raw string, backslashes pass through literally, so the resulting regex is `llo\\Z` — an escaped backslash followed by a literal `Z`. That is a *valid* pattern in the dialect, so this `.invalid` fixture no longer exercises the `\Z` end-of-string anchor that RFC 0006 excludes, and a conforming implementation should accept it — the fixture stops being invalid for the reason its name claims.

The sibling fixtures show the convention: the lower-z file uses `r"llo\z"` and the backreference file uses `r"(a)\1"` — always a single backslash. Issue #171's suggested fix is also the single-backslash form.

The line should be:

```yaml
        - print(r'{{ re_search("hello", r"llo\Z") }}')
```

(I suspect the `\\Z` came from copying the issue body's markdown *source*, where `\\Z` renders as `\Z`.)

One more note: a regression assertion that only checks the two files are no longer byte-identical passes for any byte change, including this incorrect one — worth asserting the actual pattern content (`\Z`, single backslash) instead.

Sanjays2402:
Thanks @leongdl — good catch on the double backslash, that was exactly the markdown-copy issue you suspected. Fixed: the fixture now uses the single-backslash form, matching the lower-z and backreference siblings, so it genuinely exercises the `\Z` anchor rejection again.

I also amended the commit to add the Signed-off-by trailer.

On the assertion note: I couldn't find a file-identity check in this repo's tests, and the content-level coverage already exists in openjd-rs (`reject_backslash_upper_z_end_anchor` asserts the `\Z` rejection directly). If you'd like a content assertion added somewhere specific here, happy to do it — just point me at the spot.

PR #179, opened 2026-08-26
test: Add conformance fixtures for every coercible 2023-09 field

leongdl:
## What

Conformance fixtures for **every** OpenJD 2023-09 field where the raw template and the resolved job can hold different values. Seventeen fixtures added and one strengthened, across 21 changed files. Eleven of the new fixtures plus the strengthened one are live and pass both reference implementations; six sit under `proposed/` because they fail one implementation or await a spec ruling.

The suite covers these fields well as literals and barely at all as format strings or string-form numerics. That asymmetry is the whole gap.

## Why it matters

Not hypothetical. A service reading `chunks.defaultTaskCount` out of the pre-resolution template rejected valid task-chunking jobs in production across three regions, because it saw the literal `{{Param.ChunkSize}}` where the resolved job carries an `int`. The template was valid and the resolved value was already available; the consumer read the wrong one.

Writing the fixtures found four more defects that nothing in the suite could have caught.

## Two coercion layers, and the seam between them

A field annotated `@fmtstring` carries one coercion. A parameter declared `<integer> | <intstring>` carries another. Chained, they are the real customer shape and the untested one.

Every fixture whose subject is a `@fmtstring` field therefore writes its parameter value in the string form. `chunks.defaultTaskCount` resolves from `default: '03'`, so the asserted boundaries `1-3`/`4-6`/`7-8` hold only if the string parsed to 3 **and** the format string then resolved to it. `notifyPeriodInSeconds` takes a submitted `'047'`, chaining three coercions with the reflection variable. Leading zeros are deliberate: a bare `'3'` is indistinguishable from a passed-through literal, and the fixtures forbid the un-normalized text.

The assertions are identical to the native-integer form, so this is strictly stronger at no cost.

## The `<intstring>` and `<floatstring>` class had no job coverage at all

`JobIntParameterDefinition` and `JobFloatParameterDefinition` declare `default`, `allowedValues`, `minValue` and `maxValue` as a union with a string form, in **base** 2023-09, with **no extension gate**. `IntRangeList` and `FloatRangeList` do the same for their elements.

The suite had template-level accept coverage and no job-level coverage of any string-form definition field. Across 2,417 `IntRangeList` elements and 1,044 `FloatRangeList` elements, not one used the string form.

The bound case is the sharp one: when `minValue` is the string `'010'` and the submitted value is the number `50`, the comparison must be numeric. A lexical comparison orders them differently, and nothing was checking.

## Fields no CLI prints

`cancelation.notifyPeriodInSeconds` as a format string was covered only by a fixture asserting the constant `NOTIFY_OK`, because neither `openjd run` nor `openjd summary` surfaces the field. `WrappedAction.Cancelation.NotifyPeriodInSeconds` from RFC 0008 closes it properly: the reflection variable carries the effective value after resolution, so a wrap script reads back what the job actually got.

The resolved job `name` is similar. Both CLIs print it as `Running job '<name>'` and `Job: <name>`, so it is assertable at base level without EXPR. Existing base fixtures assert a parallel substitution into a task argument instead, which proves the parameter resolved somewhere and says nothing about the name field.

## Live fixtures, all passing both CLIs

| Fixture | Asserts |
|---|---|
| `TASK_CHUNKING/jobs/default-task-count-format-string` | `defaultTaskCount` from `'03'` yields boundaries `1-3`, `4-6`, `7-8`; forbids `CHUNK:1-8` |
| `TASK_CHUNKING/jobs/target-runtime-seconds-format-string` | `targetRuntimeSeconds` from `'00'`, where 3.4.1.5 fixes scheduler behaviour so boundaries are deterministic |
| `WRAP_ACTIONS/jobs/wrap-cancelation-notify-period-fmtstring-resolved` | Resolved `notifyPeriodInSeconds` read back through reflection; forbids the default and the unresolved literal |
| `base/jobs/1.1.1--resolved-job-name-value` | The resolved `name` field itself; forbids the default-derived name |
| `base/jobs/1.1--path-default-joined-with-template-dir` | PATH default joined with the template directory |
| `base/jobs/2.3--int-param-intstring-default-resolves` | `default: '007'` resolves to `7`; forbids `COUNT:007` |
| `base/jobs/1.1--int-intstring-bounds-satisfied` and `--violation.invalid` | String `minValue`/`maxValue` compared numerically against a native submitted value |
| `base/jobs/1.1--int-intstring-allowedvalues-violation.invalid` | String `allowedValues` membership decided numerically |
| `FEATURE_BUNDLE_1/jobs/3.3.1--amount-min-format-string-resolves-to-non-number.invalid` | Resolved `min` must be numeric |
| `FEATURE_BUNDLE_1/jobs/3.3.1--amount-max-format-string-resolves-to-non-number.invalid` | Same for `max`, which is a different spec type |

## One existing fixture strengthened

`base/jobs/1.1--validation-valid-path-default-within-template-dir.test.yaml` asserted the substring `/output` against a default of `./output`. `/output` is a substring of `./output`, so an implementation that skipped the template-directory join passed it. I verified that by running it with the task hardcoded to print the un-joined value: it passed. Two `forbidden` entries close it, and the same mutant now fails.

## Proposed fixtures: four implementation defects and one spec question

Under `proposed/`, which the runner's non-recursive glob skips, so the suite stays green. Promotion is `git mv` up one directory with no edit. Each directory carries a README with the construct, the observed behaviour per implementation, and a classification.

| Fixture | Observed |
|---|---|
| `TASK_CHUNKING/.../default-task-count-...-resolves-to-zero.invalid` | Python rejects. Rust reports `passes validation checks`, runs, logs `Frame(CHUNK[INT]) = 1-1` |
| `FEATURE_BUNDLE_1/.../amount-max-...-resolves-to-zero.invalid` | `max` is `<positivefloat>`; Python rejects `0`, Rust runs. The only assertion distinguishing `max` from `min` |
| `base/.../format-string-in-anyof-...-invalid-value.invalid` | Python rejects, Rust runs |
| `base/.../format-string-in-allof-...-invalid-value.invalid` | Same, swept across four invalid values |
| `base/.../int-range-intstring-elements-normalized` and the float pair | Spec question, see below |

The first four are one defect shape: a constraint that cannot be checked at decode because the value is a format string, deferred correctly, then never re-checked after resolution. Reviewers may prefer to treat them as one issue. `#161` establishes the class with `1.1.1--resolved-name-129-chars.invalid.test.yaml`, whose comment notes "the literal is short, so only a post-resolution check catches this".

The two range-element fixtures carry a **spec question, not a defect**. `<intstring>` and `<floatstring>` are defined only as base-10 string representations, with nothing said about normalization, so `['02']` yielding `2` and yielding `02` are both defensible. The implementations disagree in opposite directions on different fields: for range elements Python preserves the literal and Rust normalizes; for a FLOAT parameter default they swap sides. And `EXPR/jobs/expr1.3.4--float-passthrough.test.yaml` already pins the verbatim reading for a default, so if the spec rules that `<floatstring>` normalizes, that landed fixture and mine cannot both be right. **These two should not merge until that is settled.**

## Verification

Every fixture run against both the Python and Rust CLIs, then mutation-checked: change the resolved value, leave the expectations, confirm it fails. **Thirteen mutants, thirteen caught.** The notifyPeriodInSeconds and job-name fixtures were additionally mutated by removing their parameter overrides, so a fixture satisfied by a template default rather than a resolved submitted value fails. `'03'` to `'abc'` fails at parse rather than passing quietly.

Slice results on this branch, both CLIs: `TASK_CHUNKING` 20/0, `FEATURE_BUNDLE_1` 56/0, `WRAP_ACTIONS` 73/0, `base` 661 passed with one pre-existing failure. `proposed/` confirmed undiscovered in every slice.

CI is green on five of six legs. The Python CLI on Windows fails 13 fixtures, and that set is byte-identical to mainline's, so it is the known tolerated set rather than drift from this change. Two fixtures are POSIX-gated and log an explicit skip on Windows.

Locally I also saw `base/job_templates/3.5--env-script-onexit-only.yaml` fail against my checked-out Rust CLI, but CI passes it, so that was version drift in my binary rather than a suite problem.

## Two defects in the suite's own machinery, not fixed here

Worth separate issues.

A non-`.invalid.` `.test.yaml` with no `expected:` block never inspects the return code in `run_job`, so a test whose run errored is reported as passing. Proven with a template referencing an undefined parameter that Rust rejects outright.

Nothing enforces that `proposed/` stays undiscovered. Several PRs of known-failing fixtures now depend on a non-recursive `glob` that no test protects; making it recursive would silently turn the suite red.

## Not covered, and why

Positive assertions on resolved host-requirement values. No CLI surfaces them, and the eight WRAP_ACTIONS reflection variables are scoped to `<Action>` fields, so reflection cannot reach them. The negative cases here are the reachable half.

Capability `name` as a format string. The spec declares `<AmountCapabilityName>` "A string" at 3.3.1.1 and does not annotate it `@fmtstring`, yet Python resolves and re-validates it while Rust rejects the template at decode. A fixture today would pin an implementation rather than the spec, so this needs a ruling first. I would raise it as a spec question.

## Relationship to other open PRs

No file-level conflict: no open PR adds or modifies any path here, and none touches `TASK_CHUNKING/` at all or adds a `hostRequirements.amounts` block anywhere.

`#174` introduces a self-asserting convention for single-task `jobs/` fixtures. These are written against mainline with plain `expected.output`; whichever merges second wants an instrumentation pass, not a merge resolution. Several fixtures here are multi-task, which `#174` excludes from self-assertion anyway.

Separately for `#174`: its rewrites of `3.6--let-in-host-requirements` and `7.3--param-in-capability` make both self-asserting while leaving their subjects unasserted. The first still asserts a parallel `print` of the let binding rather than the resolved `amounts[].min`; the second still has no `hostRequirements` block at all.


conformance-tests/2023-09/base/jobs/1.1--int-intstring-bounds-satisfied.test.yaml
```diff
@@ -0,0 +1,40 @@
...
+#   §2.3 L309  maxValue is `<integer> | <intstring>`
+#     https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md?plain=1#L309
+template:
+  specificationVersion: jobtemplate-2023-09
+  name: TestJob
+  parameterDefinitions:
+  - name: Count
+    type: INT
+    minValue: '010'
+    maxValue: '100'
```
jericht, line 24: **Missed coercion (Finding 1): no FLOAT analog of this fixture.**

`JobFloatParameterDefinition` (§2.4, L363–366) declares `minValue`/`maxValue`/`allowedValues` as `<float> | <floatstring>` — identical to INT here, no extension gate — but this PR adds the string-form numeric-comparison fixtures only for INT. There is no job-level FLOAT equivalent of `1.1--int-intstring-bounds-satisfied` / `-violation` / `-allowedvalues-violation`.

Existing FLOAT coverage does not fill it: `base/jobs/1.1--float-parameter-string-coercion-valid` coerces the submitted value and declares no bounds, and `base/job_templates/2.4--float-param-minmax-floatstring.yaml` is decode-only with bounds `"0.0"`/`"1.0"` that do not diverge lexically from a numeric compare.

This is **not** blocked by the `<floatstring>` normalization spec question the proposed range fixtures defer: numeric comparison is normalization-independent (`50.0 >= 10.0` regardless of stored form), so a FLOAT bounds/allowedValues fixture can land live exactly like this one. Suggest adding floatstring `minValue`/`maxValue` with a lexically-divergent bound (e.g. `'010.0'`) plus an `allowedValues` numeric-membership case. (A FLOAT `default`-resolves fixture should stay deferred — it does collide with `EXPR/jobs/expr1.3.4--float-passthrough`.)

conformance-tests/2023-09/FEATURE_BUNDLE_1/jobs/3.3.1--amount-min-format-string-resolves-to-non-number.invalid.test.yaml
```diff
@@ -0,0 +1,35 @@
...
+  parameterDefinitions:
+  - name: CpuMin
+    type: STRING
+    default: "abc"
+  steps:
+  - name: Step1
+    hostRequirements:
+      amounts:
+      - name: amount.worker.vcpu
+        min: "{{Param.CpuMin}}"
```
jericht, line 28: **Finding 2 (secondary): amount `min`/`max` plain string-form numeric comparison is untested.**

`amounts[].min`/`.max` (§3.3.1 L958/960) are `<nonnegativefloat> | <nonnegativefloatstring>` **and** `@fmtstring`. This fixture covers the `@fmtstring`→non-number path. The plain **string-form** path is not covered anywhere: `min`/`max` written as floatstrings where `min > max` must be caught by a **numeric** comparison (`base/job_templates/3.3.1--amount-min-greater-than-max.invalid.yaml` uses numeric literals only). Same lexical-vs-numeric class as the INT bounds fixtures. Lower severity than the FLOAT parameter-definition gap, but currently a blind spot. Suggest a `min`/`max` floatstring `min > max` invalid fixture.

jericht (commented):
Overall LGTM just a couple missing test cases

jericht:
Also windows is failing

conformance-tests/2023-09/base/jobs/proposed/3.3.2--format-string-in-anyof-resolves-to-invalid-value.invalid.test.yaml
```diff
@@ -0,0 +1,39 @@
...
+# resolved value and runs the job. The Python CLI rejects it with "Value not valid! is not a
+# valid attribute capability value." That divergence is the defect this case exists to expose;
+# it is not a broken test.
+#
+# Spec references, Template Schemas 2023-09:
+#   §3.3.2 L1014  attributes[].anyOf is @fmtstring in base 2023-09, no extension gate
+#     https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md?plain=1#L1014
+#   §3.3.2.2 L1059  `<AttributeCapabilityValue>` constraints apply AFTER the format string resolves
+#     https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md?plain=1#L1059
+template:
```
leongdl, line 19: <AttributeCapabilityValue>
A [Format String](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md#73-format-strings) subject to the following constraints:

Minimum length: 1 character.
After the format string has been resolved:
Max length: 100 characters
Unicode alphanumeric characters in the latin character set, plus the underscore and hyphen characters.
Must start with either a letter character or the underscore character.

conformance-tests/2023-09/base/jobs/proposed/3.3.2--format-string-in-allof-resolves-to-invalid-value.invalid.test.yaml
```diff
@@ -0,0 +1,42 @@
...
+#     https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md?plain=1#L1014
+#   §3.3.2.2 L1059  `<AttributeCapabilityValue>` constraints apply AFTER the format string resolves
+#     https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md?plain=1#L1059
+template:
+  specificationVersion: jobtemplate-2023-09
+  name: TestJob
+  parameterDefinitions:
+  - name: Feature
+    type: STRING
+    default: "not valid!"
```
leongdl, line 28: <AttributeCapabilityValue>
A [Format String](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md#73-format-strings) subject to the following constraints:

Minimum length: 1 character.
After the format string has been resolved:
Max length: 100 characters
Unicode alphanumeric characters in the latin character set, plus the underscore and hyphen characters.
Must start with either a letter character or the underscore character.

conformance-tests/2023-09/base/jobs/proposed/3.4.1.1--int-range-intstring-elements-normalized.test.yaml
```diff
@@ -0,0 +1,41 @@
...
+template:
+  specificationVersion: jobtemplate-2023-09
+  name: TestJob
+  steps:
+  - name: Step1
+    parameterSpace:
+      taskParameterDefinitions:
+      - name: Frame
+        type: INT
+        range: ['1', '02', '003']
```
leongdl, line 26: https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md#3411-inttaskparameterdefinition

conformance-tests/2023-09/base/jobs/proposed/3.4.1.2--float-range-floatstring-elements-normalized.test.yaml
```diff
@@ -0,0 +1,45 @@
...
+template:
+  specificationVersion: jobtemplate-2023-09
+  name: TestJob
+  steps:
+  - name: Step1
+    parameterSpace:
+      taskParameterDefinitions:
+      - name: Weight
+        type: FLOAT
+        range: ['1.5', '02.50']
```
leongdl, line 32: https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md#3412-floattaskparameterdefinition

conformance-tests/2023-09/base/jobs/1.1--int-intstring-bounds-satisfied.test.yaml
leongdl: Interesting and we should add it if the spec is missing it. This would need a new test case, not changing this one.

conformance-tests/2023-09/FEATURE_BUNDLE_1/jobs/proposed/3.3.1--amount-max-format-string-resolves-to-zero.invalid.test.yaml
```diff
@@ -0,0 +1,43 @@
...
+# unguarded.
+#
+# Spec references, Template Schemas 2023-09:
+#   §3.3.1 L959  amounts[].max is `<positivefloat>`, so 0 is illegal
+#     https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md?plain=1#L959
+#   §3.3.1 L957  amounts[].min is `<nonnegativefloat>`, so 0 IS legal there
+#     https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md?plain=1#L957
+#   §3.3.1 L960  max may be a format string under FEATURE_BUNDLE_1, so the bound defers past decode
+#     https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md?plain=1#L960
+template:
```
leongdl, line 21: https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md#331-amountrequirement

conformance-tests/2023-09/FEATURE_BUNDLE_1/jobs/3.3.1--amount-min-format-string-resolves-to-non-number.invalid.test.yaml
leongdl: Agreed, I will create more tests to cover this as well.

conformance-tests/2023-09/base/jobs/1.1--int-intstring-bounds-satisfied.test.yaml
leongdl: Added as a separate PR to not bloat this very long one - https://github.com/OpenJobDescription/openjd-specifications/pull/182/changes

PR #184, opened 2026-09-11
test: pin job-name length limits after parameter substitution

wyongzhi:
Adds three job-execution conformance fixtures that make the job name cross a length boundary through job-parameter substitution:

| Fixture | Extensions | `{{Param.N}}` resolves to | Expected |
|---|---|---|---|
| `2023-09/base/jobs/1.1.1--job-name-too-long-substitution.invalid.test.yaml` | none | 129 chars | job creation fails |
| `2023-09/FEATURE_BUNDLE_1/jobs/1.1.1--job-name-512-chars-substitution.test.yaml` | FEATURE_BUNDLE_1 | 512 chars | valid |
| `2023-09/FEATURE_BUNDLE_1/jobs/1.1.1--job-name-513-chars-substitution.invalid.test.yaml` | FEATURE_BUNDLE_1 | 513 chars | job creation fails |

### Why

The Template Schemas define the `JobName` limit as "128 characters, after the format string has been resolved. 512 characters, after the format string has been resolved if using the `FEATURE_BUNDLE_1` extension" (§1.1.1). Every existing length-boundary fixture (`1.1.1--max-length.yaml`, `1.1.1--too-long.invalid.test.yaml`, `1.1.1--job-name-512-chars.yaml`, `1.1.1--job-name-513-chars.invalid.yaml`, `1.1.1--job-name-without-extension.invalid.test.yaml`) uses a literal name, and every fixture whose name is a format string resolves to 30 characters or fewer. So an implementation that applies the limits to the template literal at decode time but mishandles the resolved value at job creation passes the suite today. These fixtures lock the "after the format string has been resolved" clause for both the base and the FEATURE_BUNDLE_1 limit. The template is short and decodes fine in every case; the outcome is decided only at job creation, which is what `.invalid.test` is for (README: "template passes static validation but the error only fires during evaluation").

The name is a bare `{{Param.N}}` so the resolved length equals the parameter length exactly; the long values are written as literal `A` runs, as in `1.1.1--job-name-512-chars.yaml`. The valid 512 fixture asserts the task's output (`OUTPUT:512`, the resolved name length): the runner only checks `expected.output` for valid job tests, so without it a job-creation failure would be reported as a pass. Each fixture carries a header comment naming the spec section and the limit it pins. STRING parameter defaults are capped at 1024 characters, so none of the three values can fail at decode for a parameter-value reason.

### Testing

- YAML loads; parameter default lengths verified as 129 / 512 / 513.
- `run_openjd_cli_tests.py` with openjd-cli 0.7.7 against the released openjd-model-for-python 0.11.11 (no companion fix): the 512 fixture fails with `Missing expected output: OUTPUT:512` and the run error `String should have at most 128 characters`; the two `.invalid.test` fixtures pass. This is the gap the fixture exists to catch.
- Same runner against openjd-model-for-python with the companion fix: 3 passed / 0 failed; the `.invalid.test` cases reject with "at most 128 characters" and "at most 512 characters", the 512 case creates, runs, and prints `OUTPUT:512`.
- The runner discovers fixtures by glob; no index to update.

### Related

Companion fix in openjd-model-for-python: https://github.com/OpenJobDescription/openjd-model-for-python/pull/362 (without it, the 512 fixture fails on the Python implementation; the Rust implementation already passes all three).

---


conformance-tests/2023-09/base/jobs/1.1.1--job-name-too-long-substitution.invalid.test.yaml
```diff
@@ -0,0 +1,16 @@
+template:
```
leongdl, line 5: Please add a description to the test what it is testing.

EG: Section x.y.z of the spec, it is invalid to have a string longer than 128 (or something like this)

conformance-tests/2023-09/FEATURE_BUNDLE_1/jobs/1.1.1--job-name-512-chars-substitution.test.yaml
```diff
@@ -0,0 +1,18 @@
+template:
+  specificationVersion: jobtemplate-2023-09
+  extensions:
+  - FEATURE_BUNDLE_1
+  name: "{{Param.N}}"
+  parameterDefinitions:
+  - name: N
+    type: STRING
+    default: "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
```
leongdl, line 13: Same here, with FEATURE_BUNDLE_1, spec x.y.z, we can have a longer string 512.

leongdl (changes requested):

conformance-tests/2023-09/FEATURE_BUNDLE_1/jobs/1.1.1--job-name-512-chars-substitution.test.yaml
wyongzhi: Added, same shape: spec 1.1.1 <JobName> with FEATURE_BUNDLE_1, max 512 after resolution, resolves to 512 and must be accepted. Also added an expected.output assertion (the task prints the resolved length) — the runner only checks expected output for valid job tests, so without it a job-creation failure would still show as a pass. The 513 fixture got the same header.

conformance-tests/2023-09/base/jobs/1.1.1--job-name-too-long-substitution.invalid.test.yaml
wyongzhi: Added. Header now reads: spec 1.1.1 <JobName>, max 128 characters checked after the format string is resolved; the name resolves to 129 via {{Param.N}}, so job creation must fail while the template itself decodes fine.

wyongzhi:
The `1.1.1--job-name-512-chars-substitution` job is expected to fail here until an openjd-model release containing OpenJobDescription/openjd-model-for-python#362 is on PyPI; this workflow installs the published CLI. It passes locally against the #362 branch. The failure is the fixture doing its job: before the output assertion was added, it reported ✓ on the same buggy version.

conformance-tests/2023-09/base/jobs/1.1.1--job-name-too-long-substitution.invalid.test.yaml
leongdl: Thanks :)

leongdl (approved):

